Issue #458, opened 2021-03-13
Wrong position for "Your current PIN is incorrect." alert


r-c-k:
## Issue description

When trying to change the PIN with an incorrect current PIN the alert message is placed under the second input field instead of the first one.

## Wallet version

v0.1.0

## System specification

macOS 11.3 Beta

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

1. Go to Settings - Security - Change PIN code
2. Try changing the PIN with an incorrect current PIN

## Expected behaviour

The alert "Your current PIN is incorrect." should appear under the input field for the current PIN.

## Actual behaviour

The alert "Your current PIN is incorrect." appears under the first input field for the new PIN.

## Errors

<img width="434" alt="Bildschirmfoto 2021-03-13 um 22 24 18" src="https://user-images.githubusercontent.com/17240275/111044978-f8695b80-844b-11eb-8a5c-05c3922b8f5f.png">

obany:
This should be fixed in the next release, please re-open if the problem persists.

PR #609, opened 2021-03-22
fix: Use pin control in settings

obany:
# Description of change

Use the proper pin control in settings instead of a password box. This introduces a smaller version of the pin control which matches the style of standard inputs.
Also add error support for pin control, and corrects where the errors appear when validating the pin values.

## Link to any relevant issues

Fixes https://github.com/iotaledger/firefly/issues/458

## Type of change

Tested on windows.

## Change checklist

image

### DIFF
--- a/packages/shared/routes/dashboard/settings/views/Security.svelte
+++ b/packages/shared/routes/dashboard/settings/views/Security.svelte
@@ -1,5 +1,5 @@
 <script lang="typescript">
-    import { Button, Checkbox, Dropdown, HR, Password, Spinner, Text } from 'shared/components'
+    import { Button, Checkbox, Dropdown, HR, Password, Pin, Spinner, Text } from 'shared/components'
     import { Electron } from 'shared/lib/electron'
     import { showAppNotification } from 'shared/lib/notifications'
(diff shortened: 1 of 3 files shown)


packages/shared/routes/dashboard/settings/views/Security.svelte
```diff
@@ -196,7 +197,7 @@
                 },
             })
         } else if (newPincode !== confirmedPincode) {
-            newPincodeError = locale('error.pincode.match')
+            confirmationPincodeError = locale('error.pincode.match')
```
begonaalvarezd, line 200: What do you think about also adding an error for the setup in pin confirmation `routes/setup/protect/views/Pin.svelte`? Right now the proceed button is just disabled when the pins dont match, but maybe it is better displaying an error?

obany: Sounds like a good idea, I think there was an issue someone raised like that. Let me add to this PR

obany: Done

begonaalvarezd (approved):

Issue #715, opened 2021-03-23
Error "Addresses start with prefix atoi" when there is a typo in the address behind "atoi"


github-i286:
## Issue description
Error "Addresses start with prefix atoi" when there is a typo in the address behind "atoi"
There is sometimes alos the the error pop-up "Somthing went wrong" with a mistyped address.

We should have different message for non-network prefix addresses and typos after the network prefix.

## Wallet version
0.3.1

## System specification
Win 10

## Is it a repeatable issue?
Y

## Steps To reproduce the issue
Try to send for exmaple to the invalid address "atoi1asdl7uasduz7asdjhasd7asdhasjdkasdhbbnasdasdad23nnnasdsadas3"

## Expected behaviour

Expected error: "Invalid address: checksum failed"

## Actual behaviour
Error "Addresses start with prefix atoi" (a user would be confused, as his address does start with the network prefix.
Or "Something went wrong", even when we know it's the address that is corrupted.

## Errors
There are no errors logged.

dinangeri:
hello, i have the same prefix address ( atoi ), when i try to send a transaction it is not possible ( invalide address ) what i have to do to solve the issue ? thanks

github-i286:
> hello, i have the same prefix address ( atoi ), when i try to send a transaction it is not possible ( invalide address ) what i have to do to solve the issue ? thanks
Not every number starts with atoi is a valid address, a checksum validates addresses to avoid copy&paste and typo errors. IOTAs sent to an invalid address can never be recovered anymore, no one will ever find the private key.
The but greport here is that the error shoudl not say "Addresses start with <network-prefix>...", as sometimes they really do, but are stil invalid. I suggest in case they really do start with atoi... but is still invalid, to report "invalid address, please check again".

The atoi... prefix is just here to avoid a copy&paste error of an valid address i.e. from testnet to be used in mainnet (starts with iota...).

dinangeri:
thanks for the information and clarification

Phyloiota:
user "dinangery" experienced the following:
directly after the first profile setup he wanted to request funds from the community, so he clicked on "receiving" and coied the shown address to the clipboard, pasted the address to discord. 
I tried to send there and got following error:
<img width="283" alt="atoi" src="https://user-images.githubusercontent.com/77154511/112296370-cf5c8c80-8ccf-11eb-80d3-f93528ff843d.png">
 logout and relogg didnt change his address, he also clicked the address refresh button, but address stayed the same
[...]

PR #740, opened 2021-03-25
fix: disable send no account

obany:
# Description of change

When no account has been selected for an internal transfer the send is disabled.
Improves bech32 error reporting

Fixes https://github.com/iotaledger/firefly/issues/715

## Type of change

Tested on windows

## Change checklist

### DIFF
--- a/packages/shared/lib/utils.ts
+++ b/packages/shared/lib/utils.ts
@@ -152,7 +152,7 @@ export const isValidHttpsUrl = (url) => {
  * Validate an address given its prefix.
  * @param prefix The bech32 hrp prefix to match.
  * @param addr The address to validate.
- * @returns True if it matches the bech32 format.
+ * @returns The error string to use if it does not validate.
  */
 export const validateBech32Address = (prefix, addr) => {
     if (!addr || !addr.startsWith(prefix)) {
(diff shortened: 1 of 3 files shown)


packages/shared/lib/utils.ts
```diff
@@ -155,7 +155,16 @@ export const isValidHttpsUrl = (url) => {
  * @returns True if it matches the bech32 format.
```
begonaalvarezd, line 155: I just realized this `@returns` is wrong, maybe we can rephrase it to something like `@returns error message string if it doesnt match the bench32 format`?

packages/shared/routes/dashboard/wallet/views/Send.svelte
```diff
@@ -259,7 +256,12 @@
     {#if !$isTransferring}
         <div class="flex flex-row justify-between px-2">
             <Button secondary classes="-mx-2 w-1/2" onClick={() => handleBackClick()}>{locale('actions.cancel')}</Button>
-            <Button classes="-mx-2 w-1/2" onClick={() => handleSendClick()}>{locale('actions.send')}</Button>
+            <Button
+                classes="-mx-2 w-1/2"
+                onClick={() => handleSendClick()}
+                disabled={selectedSendType === SEND_TYPE.INTERNAL && !to}>
```
begonaalvarezd, line 262: what about using error handlers here instead of disabling the button?

obany: Done

packages/shared/lib/utils.ts
obany: Updated

begonaalvarezd (approved):

Issue #639, opened 2021-03-22
Unable to use a comma as decimal separator


d4ytn:
## Issue description

Firefly will ignore any "," you're typing into the transaction value input field.
To support all countries with different decimal separator standards, you should translate the comma into the dot when typing.

## Wallet version

v0.3.0

## System specification

What is your operating system?

macOS Big Sur 11.2.3

## Is it a repeatable issue?

Y

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. Simulate a transfer of 0,1 KIOTA to a random address
2. You will be stuck with no separator

## Expected behaviour

I would expect to get my comma input translated into a dot by Firefly seamlessly.

## Actual behaviour

I am left with no decimal separator when using ","

## Errors

-/-

94jaku:
This should be default behaviour. Vote up.

JanOonk:
Firefly should respect regional settings set by user in the OS.

obany:
This is fixed for the next release, please re-open if you still have issues.

Travetown:
The issue is still there. If I set "German" and "EUR", then the correct comma is displayed as a point. You cannot enter a comma when sending.

Issue #704, opened 2021-03-23
Send --> Amount field is not overridable only deletable


Ephaltes:
## Issue description

When you want to send iota and paste the address in the wrong field in my example in the amount field
you can not mark the address and override it, you have to mark it ,delete it by pressing backspace 
or deleting single characters by pressing backspace but you cant for example
mark it and then press 3 to override it with the amount of 3 MIOTA

## Wallet version

0.3.1

## System specification

Win 10

## Is it a repeatable issue?

Y

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. Copy a iota address into the amount field
2. mark the content of amount field
3. try to override it by pressing e.g. 5
4. the address in the amount field doesn't get overridden

## Expected behaviour

that the address gets overridden with e.g. 5

## Actual behaviour

Nothing, address is not getting overridden it just stays marked.
only way to delete from that field is by pressing backspace

## Errors

![1](https://user-images.githubusercontent.com/13035924/112135672-532e5000-8bce-11eb-846b-e3b04c023bca.png)

obany:
This should be fixed in the next release, please re-open if it still has issues.

Issue #701, opened 2021-03-23
Send amount precision not limited based on chosen denomination


janvanhoef:
## Issue description

When entering an amount to send, there is a limit to the precision of the number. Unfortunately it is possible to input numbers with a higher precision than the theoretical limit of 1 iota. See screenshot below for example. There is only 0.999999 MI, 0.999999999999999 MI shoud not be possible. Subsequently, you don't receive an error when pressing the send button.

## Wallet version

Version 0.3.1

## System specification

Windows 10

## Is it a repeatable issue?

Y

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. Create a transaction
2. Input the amount 0.999999999999999 MI
3. Press send

## Expected behaviour

Input field precision should be limited to denomination. Conversion should occur when switching denominations from a field with a lower precision to a field with a higher precision. Field should be made empty or error should occur when moving from a denomination with more precision to a denomination with less precision.

Trying to send an impossible amount should result in an error.

## Actual behaviour

Actual precision limit in field is unclear

## Errors

![image](https://user-images.githubusercontent.com/1385666/112129021-42c6a700-8bc7-11eb-8432-c9dbb490c398.png)

adamkundrat:
duplicates/similar to #632

obany:
This should be fixed in the next version, please re-open if you still have issues

Issue #670, opened 2021-03-22
It is possible to paste text in the "amount" field when sending a transaction.


lluisin:
## Issue description

It is possible to paste text in the "amount" field when sending a transaction.
The "amount" field should only accept digits in it.

## Wallet version

0.3.0

## System specification

Win 10

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

1. Copy text from somewhere (for example: Discord)
2. Paste it in "Amount" field when sending a transaction

## Expected behaviour

Red text : "The amount appears to be an invalid number"

## Actual behaviour

Text is pasted in "Amount" field

obany:
This should be fixed in the next release, please re-open if you still have issues.

PR #742, opened 2021-03-25
fix: Get decimal separator from locale

obany:
# Description of change

Now uses the selected locale to find out the decimal separator instead of always using "."
The libs in tsconfig are updated as they otherwise don't know about FormatParts api.
Also fixes allowing overwriting value when content is at max length
Allows pasting of valid numbers only.

Fixes https://github.com/iotaledger/firefly/issues/639 https://github.com/iotaledger/firefly/issues/704 https://github.com/iotaledger/firefly/issues/670 https://github.com/iotaledger/firefly/issues/701 https://github.com/iotaledger/firefly/issues/632

## Type of change

Tested on windows in german

## Change checklist

### DIFF
--- a/packages/shared/components/inputs/Input.svelte
+++ b/packages/shared/components/inputs/Input.svelte
@@ -15,6 +15,7 @@
     export let autofocus = false
     export let submitHandler = undefined
     export let disabled = false
+    export let maxDecimals = undefined
 
     let inputElement
     let decimalSeparator = getDecimalSeparator()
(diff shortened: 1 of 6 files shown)


packages/shared/lib/i18n.ts
```diff
@@ -147,8 +147,16 @@ const setLanguage = (item) => {
     setupI18n({ withLocale: locale })
 }
 
+const getDecimalSeparator = () => {
+    const numberWithDecimalSeparator = 1.1;
+    return Intl.NumberFormat(get(appSettings).language)
```
rajivshah3, line 152: I feel like we should use the browser locale here instead so that it matches the system settings. For example, in Panama (`es-PA`) the decimal separator is `.` but in Argentina (`es-AR`) the decimal separator is `,`. Using the language code from our app (`Spanish (Latin America)` is `es-419`) would make both use `.`.

```suggestion
    return Intl.NumberFormat(getLocaleFromNavigator())
```

packages/shared/components/inputs/Input.svelte
```diff
@@ -16,27 +17,25 @@
...
         value = e.target.value
     }
 
     const onKeyPress = (e) => {
-        if (e.keyCode !== 8) {
-            const isReturn = e.keyCode === 13
-            if (isReturn && submitHandler) {
+        if (e.key !== 'Tab') {
+            const isEnter = e.key === 'Enter'
+            if (isEnter && submitHandler) {
```
rajivshah3, line 29: Nice, we should make sure we use `key` instead of `keyCode` in other places as well (I can make a PR when I have some time)

packages/shared/lib/i18n.ts
obany: I wasn't sure which was the preferred behavior, if someone had specifically picked a language from settings, if that would take priority - but since we dont support all languages you might have no choice. Will go with your suggestion for now.

packages/shared/components/inputs/Input.svelte
obany: This is a nice web page which lets you type and it gives you the equivalent keys etc https://keycode.info/
Using `key` for the separator and digits means we also support number pads

rajivshah3 (approved):

Issue #700, opened 2021-03-23
deeplinks format not compatible with ARPANET rules


HBMY289:
## Issue description

I tried posting a iota deeplink on Discord, but the final link was changed automatically by Discord and was invalidated.
Any link with more than 63 characters will be cut and a slash is automatically introduced.
This link
http://1234567890123456789012345678901234567890123456789012345678901234
will be changed to this:
http://123456789012345678901234567890123456789012345678901234567890123/4

Other services might also behave similar making FireFly deeplinks in their current form unusable.

## Wallet version

0.3

## System specification

Not OS specific

## Is it a repeatable issue?

Y

## Steps To reproduce the issue

Post any iota Firefly deeplink on Discord and check the final link.

## Expected behaviour

Posted link matches the link I have entered. Clicking the link will open Firefly and show the transaction to send.

## Actual behaviour

Discord changes the link. Clicking it will open Firefly, but it will not open any transaction to send, because the link is invalid.

## Proposed solution
User TOBY MZ analyzed the issue. This is his summary: 

Because of ARPANET rules for host names, discord and maybe other communication platforms and services will have problems interpreting firefly deeplinks. In a discord chat, the address in the current deeplink format is interpreted as DNS Label, which has a maximum length of 63. To prevent platforms from misinterpreting iota deeplinks, the format should followed the internationally applied specification and implementation of domain Names postulated in rfc1035 (1987) [https://www.ietf.org/rfc/rfc1035.txt]
A possible solution to the problem in discord would be to change the format from 
iota://address/?amount=x 
to iota://send/address/?amount=x

Other web services might also restrict the second or the third field and we might even need to divide the address.  Which is not elegant but probably the safest solution.

obany:
This will be updated in the next release with the new format

`iota://wallet/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1000`

JanOonk:
> 
> 
> This will be updated in the next release with the new format
> 
> `iota://wallet/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1000`

That format didn't work for me in v0.4 beta.
This worked: iota://atoi1qqzvmh85v36z054guuscakrxvqhnpu0hwf4989g9hjg529sqyaw9z3fhml0/?amount=1000

obany:
This PR has not been merged yet

[...]

PR #773, opened 2021-03-30
fix: New deep link format

obany:
# Description of change

The original format for deep links can no longer be supported as the host part name of a uri is limited to 63 characters and addresses are 64.

This PR introduces a new format where the `host` part of the url is now a scope for the operation (so in the future we could also support `iota://did/ ` `iota://chat/` etc

The operation being performed is supplied as the first part of the path after the host section.

e.g. `iota://wallet/send/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1000`

In addition there is now an optional `unit` parameter, if not supplied the `amt` is assumed to be `i` units

e.g. `iota://wallet/send/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1.5&unit=Mi`

Fixes https://github.com/iotaledger/firefly/issues/884 https://github.com/iotaledger/firefly/issues/700 

## Type of change

Tested on windows, only works in packaged version of the build

## Change checklist


adamkundrat:
Just of interest, the code change seems pretty standard. What's the hold up here? Are there concerns about website injecting "malicious" code within the link?

sascha1337:
passing any value from external possibly requires some solid audits, so that could be the case here.

F-Node-Karlsruhe:
> 
> 
> passing any value from external possibly requires some solid audits, so that could be the case here.

Still, this feature already existed in Trinity, so the security concerns should be sorted out already.
I'd really love to see this PR merged soon :)

FlorianWendelborn:
Additionally, a wallet that can't be used to easily pay for things (or send data in the case of IOTA) is more-or-less worthless for actual usage. As it stands now, Firefly is only good for people not actually wanting to use IOTA to send payments or data.

FlorianWendelborn:
Additionally, a wallet that can't be used to easily pay for things (or send data in the case of IOTA) is more-or-less worthless for actual usage. As it stands now, Firefly is only good for people not actually wanting to use IOTA to send payments or data.

adamkundrat:
> Additionally, a wallet that can't be used to easily pay for things (or send data in the case of IOTA) is more-or-less worthless for actual usage. As it stands now, Firefly is only good for people not actually wanting to use IOTA to send payments or data.

Almost there, see pull: https://github.com/iotaledger/firefly/pull/1244

PR #792, opened 2021-04-01
feat: Reduce scroll bar width but maintain active area

obany:
# Description of change

Reduce the visible width of the scroll bar but maintain the active area around the visible part.
Some element may need to apply the `scroll-secondary` or `scroll-tertiary` styles if they are on a different background color to the default.

## Type of change

Tested on windows in light and dark mode

## Change checklist

### DIFF
--- a/packages/desktop/App.svelte
+++ b/packages/desktop/App.svelte
@@ -6,7 +6,7 @@
     import { Electron } from 'shared/lib/electron'
     import { addError } from 'shared/lib/errors'
     import { goto } from 'shared/lib/helpers'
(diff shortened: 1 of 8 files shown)


begonaalvarezd (approved):
Would be cool in the future having a border that doesnt eat space from the scrollbar top and bottom, so the thumb can move all the way to the top and bottom. But that is for another time. Acceeeepppteedd

obany:
Yeah, I tried all sorts of -ve margins, padding, position, but the scrollbars appear to have a fixed location

Issue #721, opened 2021-03-23
adding acccount, starts sending transaction from an old account to itself.


wonkytonky2:
## Issue description

after messing around sending money internally alot and after deleten and adding accounts 

for some reason  now when i add account it takes a lot of time.  and it is doing an tx in background sending existing wallet money in another  acount to itself making a real tx on tangle. 
![image](https://user-images.githubusercontent.com/48189056/112226333-548f6500-8c2e-11eb-95b4-035a95f5c2e7.png)
![image](https://user-images.githubusercontent.com/48189056/112226497-81dc1300-8c2e-11eb-8925-b2d6b26e7cd6.png)
it pretents to send money to this new acount 5 i created but it didnt do it in reality

![image](https://user-images.githubusercontent.com/48189056/112226618-a932e000-8c2e-11eb-9a46-56f67f8ef848.png)
when trying to send money internaly it shows the same acount  3 times although there are more acounts then "account6"

## Wallet version
3.1 beta

## System specification

What is your operating system?

win 10

## Is it a repeatable issue?

in my wallet yes, i dont know how i got there
it keeps happening 

## Steps To reproduce the issue
try to import my backup

pin is 555555
http://www.filedropper.com/firefly-backup-2021-03-23t22-07-00

password : give me a name in discord i should contact

lets hope you can reproduce it this way.
## Expected behaviour

just add acount don't do wierd stuff like sending tx to itself in other wallets

cvarley100:
Hey please contact me (Charlie) on Discord with the Stronghold and password.

wonkytonky2:
update : so i created a new profile and could  reproduce it manually :   create few accounts
acount 1
acount 2 
etc..
 send money  trough them .. from 1 to  last one you created

 then remove the first account 
  then add new acount again and then the problems seem to start

obany:
This should be fixed with the next release, please re-open if the problem persists

PR #796, opened 2021-04-01
fix: Make deleted accounts a function of the UI

obany:
# Description of change

This PR aims to fix that problem that there is no realistic way of deleting an indexed account from the wallet.db without it being rediscovered on syncing.

By introducing a deletedAccounts entry in the users profile we hide from view any accounts that have been deleted. Which means the wallet processing happens as though it still exists but you don't see it in the UI.

Additionally you can enable a setting that will allow you to view deleted accounts, but not perform any operations on them. They will not appear in any lists for internal send/internal transfers/receive.  But you can still view their transaction history, graphs etc

If you import the same seed on another device the accounts will reappear (you can soon delete them again), but until we have a better solution for marking the accounts in the wallet as deleted this is the cleanest solution.

Should fix issues like https://github.com/iotaledger/firefly/issues/721

## Type of change

Tested in windows

## Change checklist

image
image
image
image


packages/shared/components/modals/AccountActions.svelte
```diff
@@ -56,10 +102,12 @@
         <HR />
         <!-- Delete -->
         <button
-            on:click={() => handleDeleteAccountClick()}
+            on:click={() => (canDelete ? handleDeleteAccountClick() : hidden ? handleShowAccountClick() : handleHideAccountClick())}
```
lucasfernog, line 105: this should be a dedicated function in the future, too many `?` and `:`

lucasfernog (approved):

PR #819, opened 2021-04-09
fix: Fixes the exception in chart when there is no longer an active profile

obany:
# Description of change

If you are in account view and you logout an exception is triggered in the chart currency formatting. This is because as the views are torn down the activeProfile is no longer available and so the currency is undefined.

## Type of change

Test on windows

## Change checklist


begonaalvarezd (approved):
👏🏼

PR #820, opened 2021-04-09
fix: Internal transfer for whole amount not confirming

obany:
# Description of change

When you transfer the whole amount between 2 accounts you only receive one confirmation from the wallet library, this is because there is actually only one message to confirm.

To remedy this instead of waiting for the second message with the same id we track the details at the initiation of the internal transfer. This means we have the details for both accounts when the message confirmation is received. 

If you logout before an internal transfer completes this tracked information will be lost, under these conditions we use the internal flag of the payload to determine it was an internal transfer and display a new message without the account names.

In addition we scan all the messages stored for the wallet to set the confirmed flag, not just the account referenced in the payload.

To avoid multiple notifications for the same transfer we check that the confirmation state for a message has actually confirmed before showing the notification.

## Type of change

Tested on windows

## Change checklist

### DIFF
--- a/packages/shared/lib/wallet.ts
+++ b/packages/shared/lib/wallet.ts
@@ -61,7 +61,13 @@ type WalletState = {
     balanceOverview: Writable<BalanceOverview>
     accounts: Writable<WalletAccount[]>
     accountsLoaded: Writable<boolean>
-    confirmedInternalMessageIds: Writable<{ [key: string]: number }>
+    internalTransfersInProgress: Writable<{
+        [key: string]: {
+            from: string
+            to: string
+            amount: number
+        }
+    }>
 }
 
 type BalanceTimestamp = {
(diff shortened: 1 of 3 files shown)


packages/shared/lib/wallet.ts
```diff
@@ -561,6 +555,26 @@ export const initialiseListeners = () => {
...
 
+const updateAllMessagesState = (accounts, messageId, confirmation) => {
+    let conirmationHasChanged = false
+
+    accounts.update((storedAccounts) => {
+        return storedAccounts.map((storedAccount) => {
+            return Object.assign<WalletAccount, Partial<WalletAccount>, Partial<WalletAccount>>({} as WalletAccount, storedAccount, {
+                messages: storedAccount.messages.map((_message: Message) => {
+                    if (_message.id === messageId) {
+                        conirmationHasChanged = _message.confirmed !== confirmation
```
cvarley100, line 566: `conirmationHasChanged` 👀

obany: my keyboard doesn't work property ;)

cvarley100 (approved):

PR #858, opened 2021-04-12
feat: Add getNodeInfo support bridge to wallet lib

obany:
# Description of change

This PR adds the necessary support in the electron bridge to the backend to call the getNodeInfo API.

## Type of change

Tested on windows

## Change checklist


lucasfernog (approved):

packages/shared/lib/validator.ts
```diff
@@ -582,6 +583,103 @@ class TypeValidator extends Validator {
...
+     * @method isValid
+     *
+     * @param {MessageResponse} response
+     *
+     * @returns {ValidationResponse}
+     */
+    isValid(response: MessageResponse): ValidationResponse {
+        const payload = response.payload as NodeInfo
+
+        if ('object' !== typeof payload.nodeinfo) {
```
rajivshah3, line 602: I'm not sure if this would ever happen in this case, but `typeof` will actually return `object` if a variable is null. Would that then result in `Cannot read property 'version' of null` later in this function?

obany: I have improved the check, as you said very unlikely to happen, but in the case we dont get a properly formatted response from a broken node or load balancer etc it should be handled.

rajivshah3 (approved):

Issue #883, opened 2021-04-14
FireFly 0.4.0 does not allow decimal point entry of Mi (German, Swiss German system settings)


github-i286:
## Issue description

Entering a decimal point for a Mi transfer is not possible.
It only allows a comma ",".
![image](https://user-images.githubusercontent.com/24428771/114662963-da687280-9cf9-11eb-8d65-633ab70dc03a.png)

However when proceeding, the comma is removed and only integer Mi are transferred.
![image](https://user-images.githubusercontent.com/24428771/114663238-45b24480-9cfa-11eb-80f3-997e4eef56ee.png)

The only way to send fractions is to switch to Ki or i units.

## Wallet version

FireFly 0.4.0, German Language
Possibly is not work also with other language settings.

## System specification

Windows 10, Swiss German Settings:
![image](https://user-images.githubusercontent.com/24428771/114663150-25828580-9cfa-11eb-9369-360a2e7ea720.png)
Possibly is not work also with other language settings.

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. Set Windows to Swiss German (decimal point '.')
2. Install FireFly 0.4.0 Wallet, set German
3. Try to transfer fractions of Mi. Entering decimal point is not possible. Entering a comma is possibly, but fractions are ignored.

## Expected behaviour

Allow fractions if Mis to be seind.

## Actual behaviour

You need to switch to Ki o i to send fractions of Mi

## Errors

No error is displayed, fractions are silently ignored.

Travetown:
In my setting / German / EUR I have set the normal standard for Germany (comma for decimal places). In this case, however, I cannot enter a comma.

<img width="357" alt="Bildschirmfoto 2021-04-14 um 09 03 20" src="https://user-images.githubusercontent.com/66588710/114667833-43eb7f80-9d00-11eb-904c-5a4480941e23.png">

obany:
I acknowledge there is a bug here with formatting, parsing.
The currency format is actually picked up from the browser/electron language (would normally match the OS) and is not affected by any of the settings in Firefly itself.

github-i286:
It would be very confusing for a end user when not to use the OS system settings.
I don't even know where I would be able to change the settings of the electron browser? I habe Egde, IE, Chrom and Firefox installed on my computer.
[...]

PR #888, opened 2021-04-14
Fix : Currency formatting

obany:
# Description of change

Currency format for display and input is now determined by the language you select in settings. It is consistently applied to every location that displays a currency.

Fixes https://github.com/iotaledger/firefly/issues/883

## Type of change

Tested on windows with both English and German which use different decimal separators.

## Change checklist


github-i286:
Just my 2 cents:
- I don't think the actual formatting depends on the selected currency (like EUR, $), etc. --> It depends on the locals settings on a computer, on how a users likes to see currency values (independent of the currently itself). 
- For currency display also consider to have 1000s separator displayed (as this is very usual). For example, locals setting define the thousand separator either as apostrophe ' or as comma , . Example: 1'234'567.89 CHF (1234567.89 CHF would be very hard to read if its a million or not).
- There are many more locals settings to consider, like how negative numbers as displayed
I recommend to use local user settings, and I'm pretty sure there standard libraries that do the job for you (?). You very easily get into hot water when trying to do the job yourself.

lucasfernog (approved):
Working great. Only "problem" I see is that the chart currency is always USD, so it might look a little different in that case.

PR #898, opened 2021-04-14
fix : Invalid balance calculation

obany:
# Description of change

When account metadata was loaded the loop exited when it reached the account index which was the same as the number of accounts. The accounts sometimes do not come from the wallet in index order, so this loop aborted early.

This fix makes sure all values are taken into account, and sorts the accounts based on their index in other location to make sure similar situations do not occur.

## Type of change

Tested on windows with a stronghold backup which had unsorted accounts.

## Change checklist

### DIFF
--- a/packages/shared/routes/dashboard/wallet/Wallet.svelte
+++ b/packages/shared/routes/dashboard/wallet/Wallet.svelte
@@ -32,7 +32,7 @@
         WalletAccount,
     } from 'shared/lib/wallet'
     import { onMount, setContext } from 'svelte'
-    import { derived, get, Readable, Writable } from 'svelte/store'
+    import { derived, Readable, Writable } from 'svelte/store'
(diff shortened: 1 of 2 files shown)


packages/shared/routes/dashboard/wallet/Wallet.svelte
```diff
@@ -120,23 +120,28 @@
...
+                                const account = prepareAccountInfo(payloadAccount, meta)
+                                newAccounts.push(account)
                             } else {
                                 console.error(err)
                             }
+
+                            completeCount++
+
+                            if (completeCount === accountsResponse.payload.length) {
+                                accounts.update((accounts) => [...accounts, ...newAccounts].sort((a, b) => a.index - b.index))
```
lucasfernog, line 141: this could probably be `accounts.update(_ => newAccounts)` but ok

lucasfernog (approved):

Issue #909, opened 2021-04-17
Fiat conversion sometimes not working


r-c-k:
## Issue description

Sometimes the fiat conversion is not working and firefly shows "NAN USD" instead.

## Wallet version

v0.4.0

## Errors

<img width="125" alt="Bildschirmfoto_2021-04-17_um_12 15 44" src="https://user-images.githubusercontent.com/17240275/115110381-72c15a00-9f7b-11eb-8204-4099d18e5646.png">

PR #914, opened 2021-04-18
fix: Market data not available causes NaN values

obany:
# Description of change

When the market data endpoint is down currency conversion displays NaN. The currency formatting PR partially resolves this by not displaying conversions if it can't perform them.

This PR keeps a cache of the last valid response from the market data endpoint and load this on startup to make sure we have at least some data.

Fixes https://github.com/iotaledger/firefly/issues/909

## Type of change

Test on windows with no connection

## Change checklist

### DIFF
--- a/packages/shared/lib/marketData.ts
+++ b/packages/shared/lib/marketData.ts
@@ -16,7 +16,7 @@ const DEFAULT_MARKETDATA_ENDPOINT_TIMEOUT = 5000
 /**
  * Default interval for polling the market data
  */
- const DEFAULT_MARKETDATA_POLL_INTERVAL = 300000 // 5 minutes
+const DEFAULT_MARKETDATA_POLL_INTERVAL = 300000 // 5 minutes
 
 export enum HistoryDataProps {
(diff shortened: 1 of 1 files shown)


begonaalvarezd:
@obany if for some reason the user is migrating and the marketdata endpoint has never worked for the user (maybe it is the first time using the wallet) so there is no `localStorage` with marketData, the issue persists because I think the bug #909 makes reference to the balances  that use `convertToFiat` (in feat/migration it is used also in the setup routes like in [Balance.svelte](https://github.com/iotaledger/firefly/blob/feat/migration/packages/shared/routes/setup/Balance.svelte) & [Migrate.svelte](https://github.com/iotaledger/firefly/blob/feat/migration/packages/shared/routes/setup/migrate/views/Migrate.svelte))

obany:
@begonaalvarezd  The currency formatting PR which has just been merged resolves that, if the currency conversion can't take place it hides the currency version of amounts.

begonaalvarezd (approved):

PR #945, opened 2021-04-20
Feat: block migration during snapshot

begonaalvarezd:
# Description of change

The purpose of this PR is to block the user to enter a migration if the snapshot is happening. 

The sensitive areas are `getMigrationData` and `sendMigrationBundle`. The idea is checking the snapshot status each time one of those functions are called, and if snapshot is happening, we display a popup blocking the UI, subscribe to github to wait until the snapshot is over, unblock the user and proceed as normal. 

**Edit: we also block dashboard now**
**Edit2: we also block login now (temp kill switch)**

image

N/A

## Type of change

To test it, you can change `CHRYSALIS_VARIABLES_ENDPOINT` in `packages/shared/lib/migration.ts` to point to a local gist or any endpoint valid for testing purposes. The response expects a json like:

```json
{
    "snapshot": false
}
```
^^ set to true to simulate a snapshot happening 

Additionally, you need to add your testing endpoint to `packages/desktop/public/index.html` `connect-src 'self'` line to allow the connection.

## Change checklist


packages/shared/lib/migration.ts
```diff
@@ -143,44 +157,52 @@ export const createMigrationBundle = (inputAddressIndexes: number[], offset: num
...
-                        if (bundle.bundleHash === bundleHash) {
-                            return Object.assign({}, bundle, { migrated: true })
-                        }
-
-                        return bundle
+    return new Promise(async (resolve, reject) => {
+        // Migration: snapshot check
+        await checkChrysalisSnapshot()
+        if (get(ongoingSnapshot) === true) {
+            // we dont display the error because a notification popup will open already
```
laumair, line 164: How come? Say the network stopped while the user was migrating multiple bundles?

packages/shared/lib/chrysalis.json
```diff
@@ -0,0 +1,3 @@
+{
```
laumair, line 1: Why do we need this JSON?

packages/shared/lib/migration.ts
begonaalvarezd: We won't display an error notification, a red toast, but we will open the popup for the user and the bundle transaction will still display an error, maybe the message is confusing

packages/shared/lib/chrysalis.json
begonaalvarezd: Its the JSON that will be used for the snapshot so it doesnt return a 404, but we can simply remove it and create it when we need it

obany: I would move it to a config folder or similar nearer the root of the repo so its more obvious.

cvarley100 (approved):

cvarley100 (approved):

PR #1008, opened 2021-04-22
fix: Handle exceptions and show a dialog

obany:
# Description of change

Shows a native window dialog for unhandled exceptions both from the Electron and renderer processes.

Catches the following errors:
* Process uncaughtException from Electron runtime
* Process unhandledRejection from Electron runtime
* window.error from preload
* window.unhandledrejection from preload
* global caught exceptions from preload
* window.error from app
* window.unhandledrejection from app

The dialog is only shown in packaged mode.

image

## Type of change

Tested on windows with deliberately triggered exceptions, including deleting index.node to simulate inability to load the native modules.

## Change checklist


packages/desktop/electron/lib/errorPreload.js
```diff
@@ -0,0 +1,19 @@
...
+        ipcRenderer.invoke('error-data').then(data => {
+            callback({
+                iconPath: './assets/logos/firefly_logo.svg',
+                version,
+                diagnostics: data.diagnostics.map((d) => `${d.label.replace('popups.diagnostics.', '')}: ${d.value}`).join('\r\n'),
+                errorType: data.errorType,
+                error: data.error
+            })
+        })
+    },
```
rajivshah3, line 15: I think you can avoid using a callback here:

```suggestion
    getData: () => {
        return ipcRenderer.invoke('error-data').then((data) => {
            return {
                iconPath: './assets/logos/firefly_logo.svg',
                version,
                diagnostics: data.diagnostics.map((d) => `${d.label.replace('popups.diagnostics.', '')}: ${d.value}`).join('\r\n'),
                errorType: data.errorType,
                error: data.error
            }
        })
    },
```

packages/desktop/public/error.html
```diff
@@ -0,0 +1,198 @@
...
+      </p>
+    </footer>
+  </div>
+  <!-- https://github.com/electron/electron/issues/2863 -->
+  <script>var exports = exports || {};</script>
+
+  <script>
+    let errorData
+
+    window.error.getData((err) => {
```
rajivshah3, line 156:
```suggestion
    window.error.getData().then((err) => {
```

packages/desktop/electron/lib/errorPreload.js
obany: Agreed, you would never guess I copied the code from the about window ;)

rajivshah3 (approved):

rajivshah3 (approved):

Issue #1015, opened 2021-04-23
Special characters in the storage path cause a white screen on Windows


Tombaaa:
## Issue description

special character in windows 10 user name and user directory (like ä in my case) prevent creating a firefly profile and token migration. Also for the creation of a firefly profile you only get a warning message - something went wrong - (see screenshot below) 
![image](https://user-images.githubusercontent.com/77152394/115851322-19e43c80-a427-11eb-9583-cd21933766e9.png)

## Wallet version

Firefly Version 1.0.0

## System specification

Windows 10

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. start Firefly, select English, click and accept Privacy Policy 
2. select light/dark theme, continue, provide Profile Name (Tom), continue
3. then the windows "Something went wrong" appears in the right lower corner and disappears after 4 seconds
4. if you hit continue again, you can either create a new wallet or migrate or restore a wallet
5. in all cases you end up with a white screen, and nothing happens again at all

## Expected behaviour

Wallet should be created, Seed should be migrated

## Actual behaviour

White screen

## Errors

![image](https://user-images.githubusercontent.com/77152394/115852029-e9e96900-a427-11eb-9953-5da90aa0ebc0.png)

Error log: 
![image](https://user-images.githubusercontent.com/77152394/115852101-ff5e9300-a427-11eb-96da-4f1f15fffe54.png)

Diagnostic
![image](https://user-images.githubusercontent.com/77152394/115852154-0e454580-a428-11eb-9e5f-d061a12698b9.png)

Using a different user on the same computer but without special character did not trigger the issue, so it is assumed that the special character, in this case the ä, did trigger the error

By the way: Firefly 0.4.0 Beta worked with out any problems

gguemues:
I have the same problem with win 10

Holen Sie sich Outlook für iOS<https://aka.ms/o0ukef>
________________________________
Von: Tombaaa ***@***.***>
Gesendet: Friday, April 23, 2021 11:39:26 AM
An: iotaledger/firefly ***@***.***>
Cc: Subscribed ***@***.***>
Betreff: [iotaledger/firefly] Can not create firefly profile - something went wrong - error message - special character in Windows 10 user name the likely cause (#1015)


Issue description

special character in windows 10 user name and user directory (like ä in my case) prevent creating a firefly profile and token migration. Also for the creation of a firefly profile you only get a warning message - something went wrong - (see screenshot below)
[...]

PR #1017, opened 2021-04-23
fix: Windows replace use of non-ascii filename chars

obany:
# Description of change

On windows profile names with non ascii characters caused an exception e.g. `userä`

This is because the windows version of RocksDB does not support the wider character set https://github.com/facebook/rocksdb/issues/3408 and so the profile storage folders which use the profile names were not accessible.

On windows the non ASCII characters will now be replaced with their sequence numbers instead for folder names e.g. `userä` = `user00e4`

An alternative fix is to do a custom build of the rust RocksDB using this flag https://github.com/facebook/rocksdb/pull/4469 which enables UTF8 support for windows filenames.

Fixes https://github.com/iotaledger/firefly/issues/1015

## Type of change

Tested on windows with profile names with non ascii characters.

## Change checklist

### DIFF
--- a/packages/shared/lib/wallet.ts
+++ b/packages/shared/lib/wallet.ts
@@ -1,6 +1,6 @@
 import { mnemonic } from 'shared/lib/app'
 import { convertToFiat, currencies, CurrencyTypes, exchangeRates, formatCurrency } from 'shared/lib/currency'
-import { stripTrailingSlash } from 'shared/lib/helpers'
+import { encodeNonASCII, stripTrailingSlash } from 'shared/lib/helpers'
 import { localize } from 'shared/lib/i18n'
 import type { PriceData } from 'shared/lib/marketData'
 import { HistoryDataProps } from 'shared/lib/marketData'
@@ -25,6 +25,7 @@ import type { Message, Payload, Transaction } from 'shared/lib/typings/message'
 import type { MigrationBundle, MigrationData, SendMigrationBundleResponse } from 'shared/lib/typings/migration'
 import { formatUnitBestMatch } from 'shared/lib/units'
 import { get, writable, Writable } from 'svelte/store'
+import { Electron } from './electron'
 import type { ClientOptions } from './typings/client'
 import type { Duration, NodeInfo, StrongholdStatus } from './typings/wallet'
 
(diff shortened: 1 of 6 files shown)


rajivshah3:
> An alternative fix is to do a custom build of the rust RocksDB using this flag facebook/rocksdb#4469 which enables UTF8 support for windows filenames.

If I understand https://github.com/rust-rocksdb/rust-rocksdb/pull/319 correctly, then shouldn't RocksDB already be built with this flag?

obany:
It would have been, but this recent PR removed it https://github.com/rust-rocksdb/rust-rocksdb/pull/493

rajivshah3:
Ah, I see. That's weird 🤔

packages/shared/lib/wallet.ts
```diff
@@ -225,8 +226,26 @@ export const api: {
...
+
+    if (os === 'win32') {
+        sep = '\\'
+        profileDirName = encodeNonASCII(profileDirName)
+    } else {
+        sep = '/'
+    }
+
+    return {
+        path: `${userDataPath}${sep}${WALLET_STORAGE_DIRECTORY}${sep}${profileDirName}`,
```
rajivshah3, line 246: Do we need to do this for the `userDataPath` too? For example if the path is `C:\Users\üsername\AppData\Roaming\Firefly`

obany: Oh yeah, actually that means this PR won't work as we don't have control over the replacing user name character, it has to be fixed with the RocksDB version :(

rajivshah3: Ah that's annoying. I've opened https://github.com/rust-rocksdb/rust-rocksdb/issues/512 to see if there was a reason that flag was removed

obany: Although was is also strange is it looks like the wallet is using v0.15 https://github.com/iotaledger/wallet.rs/blob/b68f181604824db86da0a4bef9ad6cb11f3e9949/Cargo.toml#L29
which should have included the fix https://github.com/rust-rocksdb/rust-rocksdb/blob/b91299ca2e2278cd5ed0db35f5687fb5e37118cd/librocksdb-sys/build.rs#L137

rajivshah3: I wonder if the issue is in wallet.rs or stronghold then 🤔

obany: I traced through the code and its definitely in RocksDB its failing, specifically this line https://github.com/facebook/rocksdb/blob/cc1c3ee54eace876ad18c39f931e8e5039823930/port/win/env_win.cc#L1179

obany: Managed to build a version 0.16.0 with the `ROCKSDB_WINDOWS_UTF8_FILENAMES` flag set, built wallet.rs to reference, and then used the new version with FF, and it works 🎉 

Although there was some C++ code in rocksdb that wouldn't compile with the flag set, so had to update it 😅

rajivshah3: Nice! 🎉

obany:
Closing this as it will be fixed with an updated wallet.rs PR

Issue #996, opened 2021-04-21
[UX] Capslock indicator for password creation during setup (next input right after all-uppercase seed)


Prunkton:
## Description

during the wallet setup, there is the option to enter a seed manually. Entering the seed in all lowercase will not succeed. So I assume many users will just activate the capslock key while entering the seed.
On the next step you will get distracted by seeing the value of the seed and IMHO there is quiet a chance people will just forget to deactivate caps while quickly hacking in a new password on the next password creation field. Of cause the password validation will succeed for the same reason and people could end up with a 'lost' wallet

## Motivation
I assume the complexity of the change should be quiet small but the possible lost for single users can be significant

## Requirements

some kind of visual feedback, so the user is aware of the active capslock key
![image](https://user-images.githubusercontent.com/20245018/115622398-143c0900-a2f8-11eb-895f-656e325f9ce3.png)

### alternative solution:
transform entered characters automatically in uppercase characters while typing in the seed, so there is no motivation to use caps in the first place

PR #1018, opened 2021-04-23
feat: Display a caps lock indicator on password input

obany:
# Description of change

Displays a caps lock indicator on password input to stop people incorrectly typing passwords in the wrong case.

image

Fixes https://github.com/iotaledger/firefly/issues/996 https://github.com/iotaledger/firefly/issues/1021

## Type of change

Tested on windows

## Change checklist

### DIFF
--- a/packages/shared/components/inputs/Input.svelte
+++ b/packages/shared/components/inputs/Input.svelte
@@ -1,5 +1,5 @@
 <script lang="typescript">
-    import { Error } from 'shared/components'
+    import { Error, Text } from 'shared/components'
     import { formatNumber, getAllDecimalSeparators, getDecimalSeparator, parseCurrency } from 'shared/lib/currency'
     import { onMount } from 'svelte'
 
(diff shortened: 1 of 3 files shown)


packages/shared/components/inputs/Input.svelte
```diff
@@ -19,15 +19,23 @@
...
     let allDecimalSeparators = getAllDecimalSeparators()
     let decimalSeparator = getDecimalSeparator()
+    let capsLockOn = false
+    let hasFocus = false
 
     const handleInput = (e) => {
         value = e.target.value
     }
 
+    const onKeyDown = (e) => {
```
rajivshah3, line 35: I'm not sure if this is specific to macOS, but for some reason `keydown` is not fired when caps lock is turned off (instead I think `keyup` is fired). As a result, the caps lock warning shows until the next character is pressed, which might be a minor UX issue (not a huge deal though)

obany: I have added the handler on keyup as well, both can get the modified state without a problem, so that should catch both cases.

rajivshah3 (approved):

PR #1160, opened 2021-05-18
feat: Send fiat amounts

### LINKED ISSUE #763: Sending Fiat value instead of token amount
## Description

User BernvonBrot suggested this feature. I think it can be a nice addition for the wallet.

Instead of only let the user decide if he wants to send Gi, Mi, Ki etc
it should be a possibility to send a fixed fiat value USD or EUR.
So the wallet would need to grab the latest ticker value and the user can decide to send 100USD worth of tokens to an selected address...

## Motivation

Ease of use

## Requirements
make it possible to input fiat values in the send routine...

## Open questions (optional)

Use this section to ask any questions that are related to the feature.

## Are you planning to do it yourself in a pull request?

Yes/No.

maxwellmattryan:
# Description of change

Previously, the user was only able to enter values in the IOTA denominations (i, Ki, Mi, Gi, and Ti). This PR extends the functionality of the amount input to accommodate the user's selected fiat currency, meaning that the user can now specify a fiat amount when sending a transaction (internally and externally). 

- #763
- #1168 

## Type of change

This PR has been tested on MacOS (Catalina 10.15.7) and Linux (Ubuntu 20.04).

## Change checklist

### DIFF
--- a/packages/shared/components/inputs/Amount.svelte
+++ b/packages/shared/components/inputs/Amount.svelte
@@ -3,7 +3,8 @@
     import { Input, Text } from 'shared/components'
     import {
         AvailableExchangeRates,
-        convertToFiat,
+convertFromFiat,
+                convertToFiat,
         currencies,
         CurrencyTypes,
         exchangeRates,
@@ -35,7 +36,7 @@
     let unitsButton
     let focusedItem
 
-    $: fiatAmount = amountToFiat(amount)
+    $: amountForLabel = getFormattedLabel(amount)
     $: {
         if (amount.length > 0) {
             // TODO: Handle fiat here!
@@ -44,6 +45,8 @@
                 if (rawVal > MAX_VALUE) {
                     amount = formatUnitPrecision(MAX_VALUE, unit, false)
                 }
+            } else {
+                // TODO: write case here
             }
         }
     }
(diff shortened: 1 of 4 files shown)


packages/shared/components/inputs/Amount.svelte
```diff
@@ -25,32 +27,54 @@
...
 
     const clickOutside = () => {
         dropdown = false
     }
+
     const onSelect = (index) => {
         if (amount.length > 0) {
-            amount = formatUnitPrecision(changeUnits(parseCurrency(amount), unit, Unit.i), index, false)
+            if(isFiatCurrency(index)) {
+                if(isFiatCurrency(unit)) return
```
laumair, line 65: I am not sure why we need an early return here. Is there a situation where if the provided `index` is fiat, the unit will also be a `fiat`? IIUC, this line isn't needed.

maxwellmattryan: > I am not sure why we need an early return here. Is there a situation where if the provided `index` is fiat, the unit will also be a `fiat`? IIUC, this line isn't needed.

The early return statement covers the FIAT -> FIAT case, which only happens when the user decides to not change anything and just reselects the same fiat currency. If this check isn't made, then the fiat amount gets re-calculated into fiat again, which when I tried it just kept reducing the amount. I realize now though that this could be re-written just as `amount = isFiatCurrency(unit) ? amount : amountToFiat(amount).slice(2)`...

maxwellmattryan: Just tried re-writing some of it to help make it cleaner... how does this look? 

```
const onSelect = (index) => {
    updateUnit(index)
}

const updateUnit = (_unit) => {
    if(unit === _unit) return

    convertAmount(unit, _unit)

    unit = _unit
}

const convertAmount = (fromUnit, toUnit) => {
    if(amount.length <= 0) return

    if(isFiatCurrency(toUnit)) {
        amount = amountToFiat(amount).slice(2)
    } else {
        if(isFiatCurrency(fromUnit)) {
            let amountAsI = convertFromFiat(amount, $currencies[CurrencyTypes.USD], $exchangeRates[profileCurrency])
            amount = formatUnitPrecision(amountAsI, toUnit).replace(toUnit, '')
        } else {
            amount = formatUnitPrecision(changeUnits(parseCurrency(amount), fromUnit, Unit.i), toUnit, false)
        }
    }
}
```

laumair: @maxwellmattryan This is a lot cleaner. Looks good, thank you for the revision.

laumair (approved):
@maxwellmattryan Good stuff, thank you! 💯

PR #1203, opened 2021-06-14
Fixed multiple send_message calls to work simultaneously

melatron:
# Description of change

1. Removed the `block_on` pattern in which the tokio runtime was called guarded with a `mutex`. Currently we spawn tasks for each `async` call.
2. Removed the `mutex` which prevented to make simultaneous calls.
3. Changed `wallet.rs` version, so that we can call `handle` from multiple tokio tasks simultaneously.
4. Change function arguments in `shared/lib/typings/account.ts` to match those with the wallet.rs API (specifically [this method](https://github.com/iotaledger/wallet.rs/blob/349f83074a378ca228dd86c3c975411de9b184fe/src/actor/mod.rs#L449))

## Type of change

## Change checklist


packages/backend/bindings/node/native/Cargo.toml
```diff
@@ -21,3 +21,4 @@ tokio = "1.1"
 once_cell = "1.5"
 futures = "0.3"
 serde_json = "1.0"
+log = "0.4"
```
Thoralf-M, line 24: That's not needed anymore?

packages/backend/src/lib.rs
```diff
@@ -190,35 +206,57 @@ pub(crate) struct MessageFallback {
...
+    let response = response_rx.recv().await;
+    if let Some(res) = response {
+        let msg = match serde_json::to_string(&res) {
+            Ok(msg) => msg,
+            Err(e) => serde_json::to_string(&Response::new(
+                message.id,
+                message.message,
+                ResponseType::Error(e.into()),
+            ))
+            .unwrap(),
```
Thoralf-M, line 219: Maybe also use expect here

Thoralf-M (approved):

PR #1207, opened 2021-06-17
fix: Idle lockout

maxwellmattryan:
# Description of change

The changes included in this PR are for the purpose of patching a slight security vulnerability due to how sleep mode and the idle lockout mechanism work together. 

TODO: Add URL to issue

## Type of change

## How the change has been 

The changes have been tested on MacOS, Ubuntu, and Windows 10.

## Change checklist

### DIFF
--- a/packages/shared/components/Idle.svelte
+++ b/packages/shared/components/Idle.svelte
@@ -8,7 +8,7 @@
     let timeout
     let isDestroyed = false
 
-    function handleEvent() {
+    function handleEvent(): void {
         // The events are debounced so the component can get onDestroy
         // called and be followed by a debounced handleEvent so
         // make sure the idle doesn't get triggered again when its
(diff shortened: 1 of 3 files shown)


nicole-obrien (commented):
Some minor comments on the code. Do you have the issue number that this is a fix for, so I can read more context?

packages/shared/components/Idle.svelte
```diff
@@ -8,7 +8,7 @@
     let timeout
     let isDestroyed = false
 
-    function handleEvent() {
+    function handleEvent(): void {
```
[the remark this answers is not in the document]
maxwellmattryan: Yes, will look into a better name.

maxwellmattryan:
> Some minor comments on the code. Do you have the issue number that this is a fix for, so I can read more context?

I have an issue description, but since it is considered a security vulnerability it was never filed on GitHub.

packages/shared/lib/units.ts
```diff
@@ -2,6 +2,13 @@ import { Unit } from '@iota/unit-converter'
 import Big from 'big.js'
 import { getCurrencyPosition, formatNumber } from 'shared/lib/currency'
 
+export const MILLISECONDS_PER_SECOND = 1000
+export const SECONDS_PER_MINUTE = 60
+export const MINUTES_PER_HOUR = 60
+export const HOURS_PER_DAY = 24
+export const DAYS_PER_WEEK = 7
+export const MONTHS_PER_YEAR = 12
```
nicole-obrien, line 10:
```
export const MILLISECONDS_PER_SECOND = 1000
export const SECONDS_PER_MINUTE = 60
export const MINUTES_PER_HOUR = 60
export const HOURS_PER_DAY = 24
export const DAYS_PER_WEEK = 7
export const MONTHS_PER_YEAR = 12
```
Not sure this is the best file to store these constants, this file seems to be responsible for the units of iota. Did you check out the time-constants package?

nicole-obrien: Or we make our own file for these constants to reduce dependencies.

packages/shared/lib/time.ts
```diff
@@ -0,0 +1,45 @@
+import { get, writable } from 'svelte/store'
+
+export const MILLISECONDS_PER_SECOND = 1000
```
nicole-obrien, line 3: I think that these are global constants to be used throughout the code, so I don't think that they should be in the same file as the Idle util functions.

packages/shared/lib/time.ts
```diff
@@ -0,0 +1,45 @@
...
+export const SECONDS_PER_MINUTE = 60
+export const MINUTES_PER_HOUR = 60
+export const HOURS_PER_DAY = 24
+export const DAYS_PER_WEEK = 7
+export const MONTHS_PER_YEAR = 12
+
+/**
+ * The last timestamp that the app user was active
+ */
+export const lastActiveAt = writable<Date>(new Date())
```
nicole-obrien, line 13: If we are refactoring the idle util functions into a separate file than the svelte component then it should be separate to the time constants.

nicole-obrien (commented):
Sorry to be a pain 😄

nicole-obrien:
Would it be a good idea to have the application logout when the computer sleeps or locks anyway?

packages/shared/lib/time.ts
maxwellmattryan: Hmm then I think the constants should stay where they are then, and I'll put the idle code back in the component (as that's the only place it's used anyway). Although, I think using `time-constants.ts` isn't the way either, since that creates an inconsistency with the rest of the code in `shared/lib`. 

After some type of restructure though, would be good to look at separating constants from logic (along with re-organizing enums, interfaces, functions, etc.).

packages/shared/lib/time.ts
maxwellmattryan: I agree, but until the rest of the code is like this then we should keep it consistent I think.

packages/shared/lib/time.ts
nicole-obrien: Agree on the refactor. What inconsistencies are created?

Issue #1208, opened 2021-06-17
Error on profile and wallet name re-use


maxwellmattryan:
## Issue description

Users are unable to delete a profile and create a new one with the same profile name so long as they haven't closed Firefly from the time of deletion to the time of creation. Specifically, Firefly throws an error of type `AccountAliasAlreadyExists` with a message `can't create account: account alias already exists`.

It's worth noting that changing the wallet name before deleting the profile will allow a user to create a profile of the same name, however the renamed wallet shows up in addition to the initial "Wallet 1" despite it being empty. This PR's changes make it so that upon deleting a profile, its wallet accounts are also deleted making it so re-creating the same profile would only show a newly initialized "Wallet 1".

## Wallet version

- 1.1.0
- develop

## System specification

- Ubuntu (20.04)
- MacOS (Catalina 10.15.7)
- Windows 10

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

1. Create a new profile of some specific name (e.g. `TheProfile`)
2. Delete `TheProfile`
3. Without closing the application, create another profile with the same name, `TheProfile`

## Expected behaviour

Firefly should create the profile despite having the same name since it should've been deleted beforehand.

## Actual behaviour

Firefly doesn't delete all instance of the profile thus resulting in the inability to create a profile of the same name. It deletes even the profile folder within `~/.config/Firefly/__storage__` (Ubuntu), but there is some other data not being accounted for. It's worth noting that closing the app is a workaround. 

## Errors

The following error is thrown once from the `asyncCreateAccount` method in [wallet.ts](https://github.com/iotaledger/firefly/blob/ad770a532971350800817b4b22d58a3c0ed93289/packages/shared/lib/wallet.ts#L413):
```
{
    type: "AccountAliasAlreadyExists",
    message: "can't create account: account alias already exists",
    time: 1623955679138 (or some other number)
}
```

Every error after is this error below, thrown from the `asyncStoreMnemonic` method also in [wallet.ts](https://github.com/iotaledger/firefly/blob/ad770a532971350800817b4b22d58a3c0ed93289/packages/shared/lib/wallet.ts#L348):
```
{
    type: "StrongholdError",
    message: "the mnemonic was already stored`",
    time: 1623959183578 (or some other number)
}
```

nicole-obrien:
Fixed in: https://github.com/iotaledger/firefly/pull/1209

Issue #1187, opened 2021-06-03
Re-import via recovery phrase: StrongholdError: `the mnemonic was already stored`


dnalneh:
### Briefly describe the issue.

I successfully migrated, send some transactions and then I deleted my profile from firefly. When I added my new seed again via recovery phrase (without closing the app), an error occurs at the last step when setting up the pin. The user cannot go further. Please take a look at the pictures. (btw: Adding via backup-file works fine)

### Which version of wallet are you using?

firefly version 1.1.0

### System specification

Win 10

![grafik](https://user-images.githubusercontent.com/2178069/120632992-e525c500-c469-11eb-9522-de0cf3166efa.png)

![grafik](https://user-images.githubusercontent.com/2178069/120633050-fb338580-c469-11eb-9869-f76b46b83a94.png)

### Workaround
Simply close and open the app again, then everything is working fine

maxwellmattryan:
Hey @dnalneh, whenever this error occurred were you using the same profile name as the one you deleted beforehand? I am trying to replicate the issue but am not able to. From what I've tried however, using the same name results in an `AccountAliasAlreadyExists` error initially occurring then the `mnemonic was already stored` one afterwards - does this happen for you as well?

nicole-obrien:
Fixed in https://github.com/iotaledger/firefly/pull/1209

PR #1209, opened 2021-06-18
fix: Profile deletion

maxwellmattryan:
# Description of change

- Adds wrapper functions for the underlying `api.removeAccount(...)` method that __must__ be called if ever deleting accounts from a Firefly profile
- Fixes bug where the user cannot create a profile of a certain name if they just deleted a profile with that same name (in the same session) 
- Fixes bug where the user can delete the only account in his or her Firefly profile (so long as that profile is "new" or has no transaction history)
- Fixes (Windows-only) problem with incorrect syntax for Prettier CLI in running `yarn` script
- Fixes bug on Windows 10 causing Firefly to freeze when making underlying call to `fs.rmdirSync(...)` for removing the user's Firefly profile directory

- #1187
- #1208 
- #1607 
- #1608 

## Type of change

Tested on:

- MacOS (Catalina 10.15.7)

## Change checklist


packages/desktop/App.svelte
```diff
@@ -104,9 +104,10 @@
             addError(err)
         })
 
+        cleanupInProgressProfiles()
+
```
nicole-obrien, line 108: Was this moved on purpose and is it no longer needer to `await` for a promise?

packages/shared/lib/app.ts
```diff
@@ -87,6 +87,10 @@ export const logout = (): Promise<void> =>
         const ap = get(activeProfile)
```
nicole-obrien, line 87: Should we rename this variable to _activeProfile like you have done in other places?

packages/shared/lib/profile.ts
```diff
@@ -105,6 +105,7 @@ export const disposeNewProfile = async (): Promise<void> => {
     const np = get(newProfile)
```
nicole-obrien, line 105: Could also do this one as _newProfile?

packages/shared/lib/wallet.ts
```diff
@@ -586,6 +586,28 @@ export const asyncCreateAccount = (): Promise<void> =>
...
+export const asyncRemoveWalletAccount = (accountId: string): Promise<void> =>
+    new Promise<void>((resolve, reject) => {
+        api.removeAccount(accountId, {
+            onSuccess() {
+                /**
+                 * CAUTION: If an account is successfully removed in wallet.rs then it should also be
+                 * removed in the Firefly store. This is "inefficient" (esp. for batch deletes) but it
+                 * at least ensures data integrity / consistency between Firefly and the backend.
+                 */
+                get(wallet).accounts.update((_accounts) => _accounts.filter((wa) => wa.id !== accountId))
```
nicole-obrien, line 598: I think we should avoid shortening variable names

nicole-obrien: I think that findIndex and splice would be faster than filter, but it would probably be marginal.

packages/shared/lib/wallet.ts
```diff
@@ -586,6 +586,28 @@ export const asyncCreateAccount = (): Promise<void> =>
...
+            },
+            onError(err) {
+                reject(err)
+            },
+        })
+    })
+
+export const asyncRemoveWalletAccounts = (accountIds: string[]): Promise<void[]> =>
+    Promise.all(accountIds.map((id) => asyncRemoveWalletAccount(id)))
+
```
nicole-obrien, line 610: If we are deleting a whole profile, rather than deleting all accounts separately and doing the filter on the array many times; should we have an additional function to deleteAllWalletAccounts and then we just set the array to [] or undefined and we don't need to do all this computation.

packages/shared/lib/profile.ts
```diff
@@ -198,11 +199,10 @@ export const updateProfile = (path: string, value: ValuesOf<Profile> | ValuesOf<
  *
  * @returns {void}
  */
-export const cleanupInProgressProfiles = async (): Promise<void> => {
+export const cleanupInProgressProfiles = (): void => {
```
nicole-obrien, line 202: How come we no longer want this to be async?

nicole-obrien (commented):
Added some comments, some are just questions to help myself understand the changes.

packages/desktop/App.svelte
maxwellmattryan: Yes I believe it was typed as returning a `Promise` but nothing actually required that.

packages/shared/lib/wallet.ts
maxwellmattryan: > I think we should avoid shortening variable names

I generally agree, although I personally think it's fine using it in functional methods like `map` or `filter`, unless the operation is a bit more complex / requires a few lines to write. 

> I think that findIndex and splice would be faster than filter, but it would probably be marginal.

It probably would, although I tend to stick to the functional style rather than imperative. You're right though, it doesn't really matter since both have upper bounds of `O(n)`, assuming `splice` is constant time.

packages/shared/lib/wallet.ts
maxwellmattryan: If this is added then it must be done library side on wallet.rs as an extra endpoint to `remove_accounts`, which a dev would only want to call if they were deleting profile / account manager. But if that's the case then they also need to call `delete` from the library to actually delete the storage... I think we should look at consolidating all this logic for deleting a profile into a single API call, that'd be nice.

packages/shared/lib/profile.ts
maxwellmattryan: We can put it back in; I simply removed it because there's no asynchronous code in [that function](https://github.com/iotaledger/firefly/blob/a1e554fc29b68b1314cde6358cab80cb1e1f7774/packages/shared/lib/profile.ts#L202).

packages/shared/lib/app.ts
maxwellmattryan: Yes we should 😅

packages/shared/lib/wallet.ts
nicole-obrien: Okay lets add this to click up as a refactor for the future.

nicole-obrien (approved):
Looks good to go!

nicole-obrien:
@maxwellmattryan you didn't link the issues with a key word such as 'fixes', please could you check whether we should close the relevant issues you have linked?

PR #1210, opened 2021-06-18
fix(desktop): Fix issues loading HID and USB bindings

rajivshah3:
# Description of change

Because of issues with the `bindings` package, we need to hardcode the paths to the bindings

N/A

## Type of change

Tested on macOS in dev and prod

## Change checklist

### DIFF
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -63,7 +63,7 @@
         ],
         "appId": "org.iota.firefly",
         "afterSign": "./scripts/notarize.macos.js",
-        "asar": false,
+        "asar": true,
         "protocols": [
             {
                 "name": "IOTA URL Scheme",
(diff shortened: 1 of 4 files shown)


packages/desktop/package.json
```diff
@@ -63,7 +63,7 @@
         ],
         "appId": "org.iota.firefly",
         "afterSign": "./scripts/notarize.macos.js",
-        "asar": false,
+        "asar": true,
```
MiKeF05, line 66: Is this correct? Because in the commit "82777f1fae0e339260c28308faa38b1808ed5619" it was changed to false...

begonaalvarezd: afaik it was changed by @laumair just for testing purposes

rajivshah3: We should keep it enabled as it helps get around problems posed by the Windows path length limit

Issue #1155, opened 2021-05-14
Automatic Dark Mode / Use system preferences


tooomm:
## Description

Currently, there is `Light theme` and `Dark theme` support.
There is no ability to swap automatically between them or use/synchronize with your system default.

## Motivation

- One have to decide between the both, while there is clearly no right and wrong and preference is dependant on external factors like time or environment. For a great app experience the user has to switch regularly.
- Manual changing is tedious via the settings.
- There are smarter possibilities available in basically all modern operating systems to manage night shift, themes and particularly "Dark Mode".

## Requirements

1. Introduce ~`Use System Default` option next to strict `Dark` and `Light`
2. This should use the system preference and make the app switch between the two options instead trying to manage it manually and pin a certain theme.

## Are you planning to do it yourself in a pull request?

No.

maxwellmattryan:
Hey @tooomm, thanks for your suggestion. I actually had made changes adding this into the settings, but it was before Ledger ramped up really so it's a little behind now 😅. I am going to close this issue and move it to a discussion [here](https://github.com/iotaledger/firefly/discussions/1507) 👍

PR #1211, opened 2021-06-18
feat: System default theme

maxwellmattryan:
# Description of change

This PR adds the capability for the user to use his or her system's default theme as the theme for Firefly. 

- Adds option to use the system default theme
- Adds `lib/tests/mocks/*.ts` for mocking necessary objects or functions for tests (specifically added a `matchMedia` mock for the `window` object)
- Removes some small typos / inconsistencies from `locales/ja.json`

- #1155

## Type of change

Tested on:
- Ubuntu (20.04)
- MacOS (Catalina 10.15.7)

## Change checklist

## Todo

- [ ] Determine event listener situation for Ubuntu (works on MacOS)... this affects whether or not the user needs to restart Firefly for system changes to take effect


packages/shared/lib/appSettings.ts
```diff
@@ -1,18 +1,45 @@
...
 
+/**
+ * NOTE: This media query is performed only once to help make the UX
+ * consistent across OS platforms. To be specific, it ensures that users
+ * are required to restart Firefly for system theme changes to take effect.
+ * This is because the behaviour of this media query is different among platforms,
+ * e.g. queries on MacOS result in up-to-date information whereas Linux-based platforms
+ * result in stale information.
+ */
+const isSystemInDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches
```
rajivshah3, line 35: Would the Electron [`nativeTheme.shouldUseDarkColors`](https://github.com/electron/electron/blob/v9.4.4/docs/api/native-theme.md#nativethemeshouldusedarkcolors-readonly) property work better here?

maxwellmattryan: It probably would, but I've tried to figure out how to access that from the Electron app instance. Using `require('electron')` requires me to add `@types/node` as a dependency, and I can't `import { nativeTheme } from 'electron'` to resolve correctly. If we're okay with having a `require(...)` statement then I can make those changes.

rajivshah3: Your best bet might be to call it via IPC like we did for `getOS`:

https://github.com/iotaledger/firefly/blob/4053f14653ff59e2c0b3c3beab0be0c53328adf5/packages/desktop/electron/preload.js#L179

https://github.com/iotaledger/firefly/blob/d2dda8475b7c92519beeb40894b7723f4eb8787b/packages/desktop/electron/main.js#L393-L396

Then, you can call it like this:
https://github.com/iotaledger/firefly/blob/0c07be772e84c0827379085a5a6715553ae7b375/packages/shared/components/TitleBar.svelte#L26

maxwellmattryan: I would be good to do it this way, but we should make sure that this way is mobile-friendly somehow. @amadeu2 do you know if Capacitor has a similar way to access platform configuration info, specifically to find the system's theme?

nicole-obrien: I think we have a lot of electron stuff that will need to refactor and abstract from for mobile, so I suggest being selfish to desktop for now and choosing the best way for desktop.

nicole-obrien (commented):
LGTM, just need to decide on the open comment.

nicole-obrien (approved):

packages/shared/lib/appSettings.ts
maxwellmattryan: Since the `nativeTheme` is on the `remote` object, which if we want to use it requires enabling the `remote` module. I think we decided not to do that right before I joined a few months ago.

rajivshah3: Right, but we can expose it via IPC so we don't need to use `remote`

nicole-obrien: @rajivshah3 @maxwellmattryan Should we merge this and make an issue to refactor?

nicole-obrien (approved):
Approving this so that it can be in release, and we will add an issue to refactor/improve the solution again in the future.

packages/shared/lib/appSettings.ts
rajivshah3: Sounds fine to me 👍

rajivshah3 (approved):

PR #1215, opened 2021-06-20
Feat/mobile pincode manager

amadeu2:
# Description of change

Implemented setting a pin code on Firefly Mobile with [capacitor-secure-storage-plugin](https://github.com/martinkasa/capacitor-secure-storage-plugin).

none

## Type of change

Building for browser, Capacitor 2, Android Studio 4, Xcode 11.
Tested on real devices and iOS/Android emulators.
#### in firefly/packages/mobile 
```bash
$ yarn
$ yarn build

$ yarn android:update
# or for iOS
$ yarn ios:update
```

## Change checklist


packages/mobile/App.svelte
```diff
@@ -25,19 +25,23 @@
     } from 'shared/routes'
     import { onMount } from 'svelte'
 
+    $mobile = true
```
laumair, line 28: `mobile` is of [writable](https://svelte.dev/tutorial/writable-stores) type. We should either `update` it or `set` it.

packages/mobile/lib/pincodeManager.ts
```diff
@@ -0,0 +1,51 @@
...
+const { SecureStoragePlugin } = Plugins
+
+function handle_keychain() {
+    
+    function set(key: string, pincode: string): Promise<boolean> {
+        return SecureStoragePlugin.set({ 
+            key, 
+            value: pincode 
+        }).catch(error => {
+            console.trace({ error })
```
laumair, line 13: We should explicitly throw the error here, otherwise the caller promise chain can't won't be able to detect if an error occurs here.

packages/mobile/lib/pincodeManager.ts
```diff
@@ -0,0 +1,51 @@
+import 'capacitor-secure-storage-plugin';
+import { Plugins } from "@capacitor/core"
+
+const { SecureStoragePlugin } = Plugins
+
+function handle_keychain() {
```
laumair, line 6: `handle_keychain` -> `handleKeychain` :)

packages/mobile/lib/pincodeManager.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    function remove(key: string): Promise<boolean> {
+        return SecureStoragePlugin.remove({ key })
+            .catch(error => {
+                console.trace({ error })
+        })
+    }
+
+    return {
+        /**
```
laumair, line 32: Would be good to move the code comments to the function declaration.

packages/mobile/webpack.config.js
```diff
@@ -0,0 +1,127 @@
...
+    extensions: ['.mjs', '.js', '.ts', '.svelte'],
+    mainFields: ['svelte', 'browser', 'module', 'main'],
+    fallback: {
+        path: false,
+        fs: false,
+    },
+}
+
+/// ------------------------ Output ------------------------
+console.log(path.join(__dirname, '/public'))
```
laumair, line 25: Is this intended?

packages/shared/components/OnboardingLayout.svelte
```diff
@@ -4,34 +4,62 @@
...
+                    <slot name="leftpane__action" />
+                </div>
+            </div>
+        </div>
+        <div data-label="rightpane" style={`width: 62%;`} class="bg-gray-100 dark:bg-gray-900">
+            <slot name="rightpane" />
+        </div>
     </div>
-</div>
+{/if}
```
laumair, line 65: Missing new line.

packages/shared/components/inputs/Pin.svelte
```diff
@@ -5,14 +5,16 @@
     import { Error, Icon } from 'shared/components'
 
     const dispatch = createEventDispatcher()
+    const isWide = () => window.matchMedia('(max-width: 360px)').matches ? true : false
```
laumair, line 8: I don't think we need the ternary here. `window.matchMedia('(max-width: 360px)').matches` returns a boolean value.

packages/shared/components/inputs/Pin.svelte
```diff
@@ -82,6 +84,23 @@
...
 
+     /**
+     * for android mobile we need both onkeydown and oninput 
+     * event listeners to the input and handle the old and the new value.
+     * the auto-suggest feature or other event might follow 
+     * the keydown event and invalidate it. 
+     * @todo on ios its not needed, so we can detect ios too and disable it.
+    */
+    const changeHandlerHelper = (e, i) => {
+        //e.preventDefault
```
laumair, line 95: Would suggest removing this if not needed.

packages/shared/lib/profile.ts
```diff
@@ -76,7 +76,7 @@ export const activeProfile: Readable<Profile | undefined> = derived(
 )
 
 activeProfileId.subscribe((profileId) => {
-    Electron.updateActiveProfile(profileId)
+    //Electron.updateActiveProfile(profileId)
```
laumair, line 79: Why did we comment this out?

packages/shared/routes/setup/protect/Protect.svelte
```diff
@@ -64,7 +63,19 @@
                         throw new Error('Invalid pin code!')
                     }
 
-                    await Electron.PincodeManager.set(get(activeProfile)?.id, pin)
+                    if (mobile) {
+                        const { PincodeManager } = await import('Mobile/lib/pincodeManager')
```
laumair, line 67: I am a bit skeptical about this approach. This would create a lot of extra code in many other places. For now it is fine but the ideal way (imo) is to expose a common interface for stuff like this for mobile and desktop.

packages/shared/routes/setup/protect/Protect.svelte
```diff
@@ -64,7 +63,19 @@
...
+                        const saved = await PincodeManager.set(get(activeProfile)?.id || 'test', pin)
+                        //
```
laumair, line 69: Kindly remove this.

packages/shared/routes/setup/protect/Protect.svelte
```diff
@@ -64,7 +63,19 @@
                         throw new Error('Invalid pin code!')
                     }
 
-                    await Electron.PincodeManager.set(get(activeProfile)?.id, pin)
+                    if (mobile) {
+                        const { PincodeManager } = await import('Mobile/lib/pincodeManager')
+                        const saved = await PincodeManager.set(get(activeProfile)?.id || 'test', pin)
+                        // 
+                        console.log({ saved })
```
laumair, line 70: Kindly remove this.

laumair:
@amadeu2 
Screenshot 2021-06-29 at 1 56 07 PM
Not sure why but the back button here does not work.

packages/mobile/lib/pincodeManager.ts
amadeu2: 👌

packages/mobile/lib/pincodeManager.ts
amadeu2: That's good point, sorry I've just let this here for testing while writing, I have had to delete it. To build on that I've saw that the code of `pincodeManager.js` and `keychain.js` on desktop folder just returns the promise, since erros are handled by module itself. Do you think I can make the same as in desktop and just return the promise?

packages/mobile/lib/pincodeManager.ts
amadeu2: it's look weird to me too, that's because consumers can see this comments only when it's put on external closures functions

packages/mobile/.gitignore
```diff
@@ -0,0 +1,2 @@
+public/assets
+public/*.js
```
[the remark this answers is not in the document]
amadeu2: fixed thanks

packages/mobile/App.svelte
amadeu2: I will use set or update if to you is more readable. Also since the "mobile" store is auto-subscribed in the html code, Svelte allows to use the $ prefix in the <script> tag also for assignments, for the sake of readability I think.
The only difference is on compiled code where Svelte import set_store_value instead of simply use mobile.set

packages/mobile/webpack.config.js
amadeu2: 😬 I should have deleted it

packages/shared/components/OnboardingLayout.svelte
amadeu2: fixed 😬

packages/shared/components/inputs/Pin.svelte
amadeu2: agree, it's redundant, I'll fix

PR #1217, opened 2021-06-20
Refactor/optimize local dev

amadeu2:
Webpack / Node process and browser processes eat up my RAM / CPU and often hang. So I wanted to try esbuild and Tailwindcss-jit.
As a result, build time went from 65 seconds with ts-loader to 24 seconds with esbuild-loader and Tailwindcss-jit. In addition, the processes consume less resources and the initial load in the browsers was faster.

none

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

Building for browser, Capacitor 2, Android Studio 4, Xcode 11.
Tested on real devices and iOS/Android emulators

## Change checklist


packages/mobile/public/index.html
```diff
@@ -7,9 +7,9 @@
         <title>IOTA desktop app</title>
 
         <link rel="icon" type="image/png" href="/favicon.png" />
-        <link rel="stylesheet" href="/build/bundle.css" />
+        <link rel="stylesheet" href="/build/index.css" />
```
ingokpp, line 10: Optionaly add types to the link and script tags like in the favicon.
`<link rel="stylesheet" type="text/css" href="/build/index.css" /> `

packages/mobile/public/index.html
```diff
@@ -7,9 +7,9 @@
...
 
-        <script defer src="/build/bundle.js"></script>
+        <script defer src="/build/index.js"></script>
```
ingokpp, line 12: `<script type="text/javascript" defer src="/build/index.js"></script>`

packages/mobile/public/index.html
amadeu2: Thanks for your suggestion, but I can't see your point in this case, it seems redundant to me. There are different approaches related to defining defaults, but I'm reluctant to add optional values to any code, since the less we write the less error-prone. IMHO knowing the defaults and using / defining them into our own code is key.

WHATWG specs define the default type for resources given by the stylesheet keyword is text/css.
https://html.spec.whatwg.org/multipage/links.html#link-type-stylesheet

Maybe I'm loosing something, what you think could be gained adding optional type prop in this case?

packages/mobile/public/index.html
amadeu2: Same here, WHATWG specs says: authors to should omit the type attribute instead of redundantly setting it.
https://html.spec.whatwg.org/multipage/scripting.html#attr-script-type

Also text/javascript it's obsolete nowdays, correct one would be application/javascript
https://www.iana.org/assignments/media-types/text/javascript

PR #1231, opened 2021-06-29
feat: Account index validator

maxwellmattryan:
# Description of change

Adds validation for account index in the Ledger migration.

None

## Type of change

Tested on (w/ Ledger Nano S): 

- Ubuntu (20.04)
- MacOS (Catalina 10.15.70

## Change checklist


packages/shared/routes/setup/ledger/views/AccountIndex.svelte
```diff
@@ -9,11 +9,20 @@
     export let locale
     export let mobile
 
+    let expert = false
     let loading = false
 
+    let min = 0
+    let max = 2147483647
```
laumair, line 16: @maxwellmattryan Why this number?

laumair (approved):

packages/shared/routes/setup/ledger/views/AccountIndex.svelte
maxwellmattryan: I believe this is what the max index was in Trinity.

packages/shared/routes/setup/ledger/views/AccountIndex.svelte
```diff
@@ -25,6 +34,25 @@
...
+            n = Math.min(Math.max(n, min), max)
+
+        return n
+    }
+
+    function isValidNumber(n: number): boolean {
+        return isPositiveInteger(n) && isWithinRange(n)
+    }
+
+    function isPositiveInteger(n: number): boolean {
```
begonaalvarezd, line 48: @maxwellmattryan maybe we could add some error messages for the validators? I pasted for example `194909e` into the input and its not allowing me to continue but there is no feedback.
image


Maybe we could use something like we have in `ImportTextfield.svelte`  

image

begonaalvarezd: If we add these validators, I think we can remove the `validate` function in `packages/shared/components/inputs/Number.svelte` which forces a min/max

Issue #884, opened 2021-04-14
Clicking deeplink when Firefly is closed doesn't work properly


JanOonk:
## Issue description

Clicking a deeplink when Firefly is closed does not work. It will start FF, but after you click a profile and pin only startscreen appears.
Clicking a deeplink when Firefly is open does work: it will bring you to transaction screen which is prefilled with address and optionally amount when given

## Wallet version

v0.4

## System specification

Windows 10

## Is it a repeatable issue?

Yes

## Steps To reproduce the issue

Explain how the maintainer can reproduce the issue.

1. Make a deeplink on a webpage for example: `iota://atoi1qqzvmh85v36z054guuscakrxvqhnpu0hwf4989g9hjg529sqyaw9z3fhml0/?amount=1000`
2. Be sure Firefly is not started
3. Click deeplink
4. Firefly will start and after clicking profile and pin you will see the startscreen not the (prefilled) transaction screen

## Expected behaviour

Firefly should start and after selecting profile and typing pin you should see a prefilled transaction screen.

## Actual behaviour

Firefly is started, and after clicking profile and typing pin, you see the startscreen not the (prefilled) transaction screen

## Errors

No actual errors

obany:
This PR has not been merged yet, will appear in a future build https://github.com/iotaledger/firefly/pull/773

cvarley100:
Deep linking has been disabled until a later version.

PR #1244, opened 2021-07-06
feat: add deep links

Evanfeenstra:
# Description of change

Merge @obany's work on deeplinks, and fix the actual display of the `wallet/send` deeplink in the UI

Fixes issues #884 #700 

## Type of change

Tested on Mac, works on dev or built

## Change checklist

- [x ] I have followed the contribution guidelines for this project
- [x ] I have performed a self-review of my own code


packages/shared/routes/dashboard/Dashboard.svelte
```diff
@@ -137,6 +109,39 @@
...
+            showSystemNotification({ type: 'info', message: locale('notifications.deepLinkingIsNotEnabled') })
+        } else {
+            if ($accounts && $accounts.length > 0) {
+                let addressPrefix = $accounts[0].depositAddress.split('1')[0]
+
+                const parsedData = parseDeepLink(addressPrefix, data)
+
+                if (parsedData && parsedData.context === 'wallet' && parsedData.operation === 'send') {
+                    _redirect(Tabs.Wallet)
+                    console.log(parsedData)
```
adamkundrat, line 133: It might be a good idea to remove the console.log.

packages/shared/lib/utils.ts
```diff
@@ -47,77 +45,91 @@ export const generateRandomId = (): string => {
...
-            result.address = parsed.address
-        } else {
-            return null
-        }
+    const unitParam: Unit = url.searchParams.get('unit') as Unit ?? Unit.i
+    if (!Object.values(Unit).includes(unitParam)) {
+        addError({ time: Date.now(), type: "deepLink", message: `Unit is not recognised '${unitParam}'` })
+    }
 
-        if (parsed.message && typeof parsed.message === 'string') {
```
adamkundrat, line 109: wouldn't there be maxLength for the message we might want to validate?

adamkundrat (commented):
This is awesome work @Evanfeenstra, I just added few of my comments to help this move along. I hope nobody minds:)

packages/shared/routes/dashboard/Dashboard.svelte
Evanfeenstra: good catch! Its removed

adamkundrat:
Hi Guys,

So I was able to do some testing on this. For me the deep link seems to be working fine except units. For some reason units are always default to "Mi" even if I pass unit or I don't.

Try following URLs:
- iota://wallet/send/iota1qrl8ldyshmvzdhjc88fmjpqkeqpylyhw03tu3gx7la0ftn3vuxp0c529f68?amt=500000 - I would expect "i" to be set
- iota://wallet/send/iota1qrl8ldyshmvzdhjc88fmjpqkeqpylyhw03tu3gx7la0ftn3vuxp0c529f68?amt=5&unit=Gi - I would expect Gi to be set

Additionally, I added few comments on other parts of the code.

Thanks
Adam

packages/shared/lib/utils.ts
adamkundrat: ok just to add to this. I'm not expert on this but do we have to worry about XSS attack?

packages/shared/lib/utils.ts
```diff
@@ -47,77 +45,91 @@ export const generateRandomId = (): string => {
...
+        return
+    }
 
-        if (parsed.address.match(VALID_MAINNET_ADDRESS) || parsed.address.match(VALID_DEVNET_ADDRESS)) {
-            result.address = parsed.address
-        } else {
-            return null
-        }
+    const unitParam: Unit = url.searchParams.get('unit') as Unit ?? Unit.i
+    if (!Object.values(Unit).includes(unitParam)) {
```
adamkundrat, line 114: it might be good idea to make this case insensitive.

packages/shared/lib/app.ts
```diff
@@ -29,16 +30,17 @@ export const mnemonic = writable<Array<string>>(null)
 
 interface SendParams {
     amount: number
+    unit?: Unit
     address: string
     message: string
```
adamkundrat, line 35: I'm not sure we want to include this in the scope but I would love the idea we do.

Messages can be passed in the URL already although the Send.svetle component is not supporting it. It would make sense as part of this change we add it there as well. 

Unfortunately I can't allocate anymore time to this....but I thought it would be good to point this out.

adamkundrat (commented):
few more comments based on my testing

packages/shared/lib/app.ts
F-Node-Karlsruhe: Does that mean that the message can be passed and gets included in the transaction even though it is not displayed on the UI? That wouldn't be a big issue as most useres don't really send data with firefly manually imo

adamkundrat: yeah I don't think that's happening, messages are just dropped atm...see Send.svetle:

`const triggerSend = (internal) => {
        closePopup()
        if (internal) {
            // We pass the original selectedSendType in case we are masquerading as
            // an internal transfer by a send to an address in one of our
            // other accounts. When the transfer completes it resets
            // the send params to where it was
            internalTransfer(from.id, to.id, amountRaw, selectedSendType === SEND_TYPE.INTERNAL)
        } else {
            send(from.id, address, amountRaw)
        }
    }
`

so to support it: 
- new ui component needs to be added 
- above function needs to be updated to pass message as parameter. Does send supports it?

I just didn't spent time figuring out how **send** function works. It seems to be external.

Evanfeenstra: I think including `message` is outside of the scope of this merge. It may even need some changes in the node bindings, or even wallet.rs itself?

Evanfeenstra:
> Hi Guys,
> 
> So I was able to do some testing on this. For me the deep link seems to be working fine except units. For some reason units are always default to "Mi" even if I pass unit or I don't.
> 
> Try following URLs:
> 
> * iota://wallet/send/iota1qrl8ldyshmvzdhjc88fmjpqkeqpylyhw03tu3gx7la0ftn3vuxp0c529f68?amt=500000 - I would expect "i" to be set
> * iota://wallet/send/iota1qrl8ldyshmvzdhjc88fmjpqkeqpylyhw03tu3gx7la0ftn3vuxp0c529f68?amt=5&unit=Gi - I would expect Gi to be set
> 
> Additionally, I added few comments on other parts of the code.
> 
> Thanks
> Adam

Thanks Adam, units are fixed now

sascha1337:
Any points of review left to take a look at ? If not, lets get this merged for some usecases

Evanfeenstra:
> Any points of review left to take a look at ? If not, lets get this merged for some usecases

Its good to merge now

packages/shared/lib/app.ts
F-Node-Karlsruhe: Wallet.rs as well as the binding already supports data attachments to messages in the `TransferOptions` param of the send function.
I hope this issue will transalte in a PR anytime soon: https://github.com/iotaledger/firefly/issues/1256

FlorianWendelborn:
I’d suggest dropping `unit`, because all that does is make the API more complex. This can easily be handled in a utility function on whatever app implements this. KISS applies here IMO

adamkundrat:
> I’d suggest dropping `unit`, because all that does is make the API more complex. This can easily be handled in a utility function on whatever app implements this. KISS applies here IMO

I would agree with that. It's unnecessary. Even in our implementation of iota-button we just simply always convert everything to "i" units for the deep link.

Evanfeenstra:
> @FlorianWendelborn
> 
> > I’d suggest dropping `unit`, because all that does is make the API more complex. This can easily be handled in a utility function on whatever app implements this. KISS applies here IMO
> 
> I would agree with that. It's unnecessary. Even in our implementation of iota-button we just simply always convert everything to "i" units for the deep link.

The `unit` thing is very simple, its just two lines of code. I personally don't see any reason to remove it.

FlorianWendelborn:
@Evanfeenstra needs good unit-testing then I guess. Imagine if that part suddenly behaves wrong and you transfer 1 Gi instead of 1i

Just don’t really see the point in having it if everything’s already possible without

Evanfeenstra:
> @Evanfeenstra needs good unit-testing then I guess. Imagine if that part suddenly behaves wrong and you transfer 1 Gi instead of 1i
> 
> Just don’t really see the point in having it if everything’s already possible without

The only thing that the `unit` param does is populate the "unit" dropdown in the UI. The UI then is what creates the actual tx. 

I can imagine a use case for specifying units, for example an NFT exchange that values everything in Ki. If something costs 1.1Mi, maybe the application would like to display it at 1100 Ki to keep everything consistent? I know its something of an edge case, but I do not see any downside to it

FlorianWendelborn:
@Evanfeenstra so the wallet app will display the unit passed via the code? That’s could actually be pretty concerning.

What if I have a website where I pretend to charge 1 Mi but then I actually charge 1 Gi? A user that’s not extra careful might fall for this, especially if I only do it after they already bought 3 other things so they trust the website in that example.

sascha1337:
> @Evanfeenstra so the wallet app will display the unit passed via the code? That’s could actually be pretty concerning.
> 
> What if I have a website where I pretend to charge 1 Mi but then I actually charge 1 Gi? A user that’s not extra careful might fall for this, especially if I only do it after they already bought 3 other things so they trust the website in that example.

UIX wise, the confirmation dialog could show the shortform plus the full amount / next smaller unit in full digits - and/or the fiat value converted to the users set default currency, which will be clearly help to identify mistakes or manipulation

F-Node-Karlsruhe:
> 
> 
> > Any points of review left to take a look at ? If not, lets get this merged for some usecases
> 
> Its good to merge now

But when? :/ This feature is really useful for all kinds of usecases. I hope there is not too much to fix in the ledger update of Firefly so this one does not get left behind.

rajivshah3:
Thanks for your contributions to this, everyone! I apologize for the delay in reviewing this PR -- we were busy with the Ledger work. Unfortunately I can't give an ETA for when this will be reviewed and tested (since we updated Electron, we'll need to take another look), but we appreciate the PR 😃

Evanfeenstra:
dang, looks like since merging the Ledger stuff into the `merge-deeplinks` branch, `window.Electron` is undefined... any Firefly devs have an idea how to fix that?

cvarley100 (approved):

packages/shared/lib/app.ts
```diff
@@ -30,16 +31,17 @@ export const mnemonic = writable<Array<string>>(null)
...
     isInternal: boolean
 }
 
 /**
- * Input paramaters for sending transactions
+ * Input parameters for sending transactions
  */
-export const sendParams = writable<SendParams>({ amount: 0, address: '', message: '', isInternal: false })
-export const clearSendParams = (isInternal = false) => sendParams.set({ amount: 0, address: '', message: '', isInternal })
+export const sendParams = writable<SendParams>({ amount: 0, unit: undefined, address: '', message: '', isInternal: false })
```
maxwellmattryan, line 43: Is it best to leave this as `undefined` or should we use `Unit.i` or `Unit.Mi` here?

packages/shared/lib/deepLinking.ts
```diff
@@ -1,6 +1,91 @@
...
+import { writable } from 'svelte/store';
 
 /**
  * Indicates that a deep link request is active
  */
 export const deepLinkRequestActive = writable<boolean>(false)
+
+/**
+ * Parse a deep link for the app (iota://)
+ * @param {string} data Deep link data
```
maxwellmattryan, line 12: I think this needs just a little more TSDoc comment info for the function signature

packages/shared/locales/en.json
```diff
@@ -683,7 +683,7 @@
         "syncAll": "Sync all",
         "export": "Export",
         "exportNewStronghold": "Export a new Stronghold",
-        "enableDeepLinks": "Enable deep links",
+        "enableDeepLinks": "Enable IOTA links",
```
rajivshah3, line 686: @cvarley100 tbh, I think people maybe more familiar with the term "deep links". If we don't want to use "deep links" we should at least describe them as "`iota://` links". "IOTA links" sounds vague imo

cvarley100: The intention is that by calling them IOTA links in Firefly the terminology would become engrained.

cvarley100: You wouldn't say `https://` link you would say https link. And `deep links` is not familiar terminology for most users. Whatever we do, it involves teaching users a new phrase - so I thought why not coin a new one?

rajivshah3: > Whatever we do, it involves teaching users a new phrase - so I thought why not coin a new one?

That's the thing, though — the term "deep linking" is common enough to warrant having a [Wikipedia page](https://en.wikipedia.org/wiki/Mobile_deep_linking). I don't think we should coin a new term when one already exists and is widely accepted

PR #1281, opened 2021-07-21
hotfix: keep track of spent addresses during (ledger) migration

laumair:
# Description of change

Instead of just checking the inputs returned by wallet library for ledger migration data, also check spend statuses and balance.

N/A

## Type of change

Manually tested macOS

## Change checklist

### DIFF
--- a/packages/shared/lib/migration.ts
+++ b/packages/shared/lib/migration.ts
@@ -122,7 +122,7 @@ export const migration = writable<MigrationState>({
     data: writable<MigrationData>({
         lastCheckedAddressIndex: 0,
         balance: 0,
-        inputs: []
+        inputs: [],
     }),
     seed: writable<string>(null),
     bundles: writable<Bundle[]>([])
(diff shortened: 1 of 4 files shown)


packages/shared/lib/migration.ts
```diff
@@ -122,7 +122,7 @@ export const migration = writable<MigrationState>({
     data: writable<MigrationData>({
         lastCheckedAddressIndex: 0,
         balance: 0,
-        inputs: []
+        inputs: [],
```
Thoralf-M, line 125: Don't you need spentAddresses also here?

laumair: No, it is only needed for checking if we need to generate the next batch of addresses. So, I decided to not store it in svelte store. Also kept it optional in the types.

Thoralf-M (approved):

cvarley100 (approved):

PR #1323, opened 2021-07-29
fix: ensure correct address display for mined bundles

cvarley100:
# Description of change

- Bundle mining strips the checksum from the TRANSFER... address
- 9 further characters were being stripped again to leave an address length of 72 and yielding incorrect addresses and checksums displayed in ledger migration

## Type of change

Tested on Mac with mined and unmined bundles

## Change checklist

### DIFF
--- a/packages/shared/lib/ledger.ts
+++ b/packages/shared/lib/ledger.ts
@@ -19,6 +19,7 @@ import type { NotificationType } from './typings/notification'
 
 
 const LEDGER_STATUS_POLL_INTERVAL_ON_DISCONNECT = 1500
+const LEGACY_ADDRESS_WITH_CHECKSUM_LENGTH = 90
 
 let polling = false
 let intervalTimer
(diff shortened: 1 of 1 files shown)


packages/shared/lib/ledger.ts
```diff
@@ -243,7 +244,7 @@ export function getLegacyErrorMessage(error: any, shouldLocalize: boolean = fals
 }
 
 export function formatAddressForLedger(address: string, removeChecksum: boolean = false): string {
-    if (removeChecksum) {
+    if (address.length === LEGACY_ADDRESS_WITH_CHECKSUM_LENGTH && removeChecksum) {
```
maxwellmattryan, line 247: This might affect non-legacy transactions in the future if for whatever reason we try to remove its checksum.

cvarley100: non-legacy transactions don't have an address length of 90 - it's 64

maxwellmattryan (approved):
Changes look good!

PR #1399, opened 2021-08-13
chore: Testing

maxwellmattryan:
# Description of change

This PR integrates [Jest](https://jestjs.io/) for testing the Typescript files in `packages/shared/lib`, which contains mostly business logic for all platforms of Firefly. This will be important for whenever we would like to add new functionality, refactor existing functionality, etc.

I tried integrating [Mocha](https://mochajs.org/) as well, but was not able to successfully setup a mocked DOM environment for testing (despite trying [jsdom](https://github.com/jsdom/jsdom)). For this reason, Jest was chosen as it ships with jsdom making it quite easy to setup.

None

## Type of change

Tested on: 

- Ubuntu (20.04)

## Change checklist

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,0 +1,53 @@
+name: pr-checks
+
+on:
+  push:
+    branches:
+      - chore/**
+      - fix/**
+      - feat/**
+      - hotfix/**
+
+jobs:
+  lint:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Linter (Backend)
+        run: yarn lint:backend
+
+      - name: Linter (Desktop)
+        run: yarn lint:desktop
+
+      - name: Linter (Mobile)
+        run: yarn lint:mobile
+
+      - name: Linter (Shared)
+        run: yarn lint:shared
+
+  test:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js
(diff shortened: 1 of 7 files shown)


packages/shared/lib/profile.ts
```diff
@@ -104,7 +104,7 @@ export const activeProfile: Readable<Profile | undefined> = derived(
 )
 
 activeProfileId.subscribe((profileId) => {
-    Electron.updateActiveProfile(profileId)
+    Electron?.updateActiveProfile(profileId)
```
maxwellmattryan, line 107: This change exists because it was throwing errors of reading `updateActiveProfile` of `undefined`.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,28 @@
+name: pr-checks
```
rajivshah3, line 1:
```suggestion
name: PR Checks
```

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  push:
+    branches:
+      - chore/**
+      - fix/**
+      - feat/**
+      - hotfix/**
```
rajivshah3, line 9: If these are for PRs it might be best to just do this:

```suggestion
  pull_request:
```

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  test:
```
rajivshah3, line 12:
```suggestion
  test-shared:
```

.github/workflows/pr-checks.yml
maxwellmattryan: It mainly is just for PRs, but if you just do the `pull_request` line doesn't it only do it on PR creation?

rajivshah3: Nope, it does it on each subsequent push as well

PR #1448, opened 2021-09-09
feat: Add miota to portfolio chart

### LINKED ISSUE #718: Portfolio timeline display in currency MIOTA
## Description
The portfolio can be displaed in a timeline for USD, BTC, etc. It woudl be usefull to display history in MIOTA "currency", how much net deposit and withdraws has been done in view of MIOTAs.

## Motivation
View net add and spends over time in MIOTAs

## Requirements

Add MIOTA to the list of currences.

## Open questions (optional)
N/A

## Are you planning to do it yourself in a pull request?
No.

nicole-obrien:
# Description of change

- Added iota token holding chart to the profile dashboard charts, with a stepped line chart and a dropdown for selecting the token.
- Added iota token holding chart to the selected wallet dashboard charts, with a stepped line chart and a dropdown for selecting the token.
- Added a dropdown on the selected wallet dashboard charts for selecting the chart type, as this chart is smaller than the one on the profile.
- Updated props and styles for the dropdown component, so that you can: specify the text style of the value and labels; optionally remove the border of a unselected dropdown.

- Closes #718

## Type of change

Manually tested on:
- MacOS (Monterey 12.0.0)

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code


packages/shared/lib/chart.ts
```diff
@@ -31,9 +32,16 @@ interface ActivityTimeframe {
 }
 
 export enum DashboardChartType {
+    HOLDINGS = 'holdings',
     PORTFOLIO = 'portoflio',
     TOKEN = 'token',
 }
+
+export enum WalletChartType {
```
maxwellmattryan, line 40: Hmm is there a reason these are two separate `ChartType`s? It slightly confusing since `selectedWalletChart` and `selectedDashboardChart` use both the `DashboardChartType` type.

nicole-obrien: Ooops, `selectedWalletChart` was supposed to use `WalletChartType` not `DashboardChartType`. Created a new enum because the two dashboards have different lists of charts that can be displayed.

nicole-obrien: Updated to use the correct enum in the store.

maxwellmattryan (commented):
Hey @nicole-obrien, nice change for the wallet charts. I noticed that whenever the IOTA denomination is selected the y-axis steps display with numbers and units whereas the others display with just numbers. I personally like the numbers with the units as it's more clear, might be worth changing the other charts? Alternatively if you want it without the numbers, we'd have to indicate to users what iota denomination it's in... dropdown or `formatUnitBestMatch()` might be helpful.

I am wondering if it makes sense to put IOTA into the "Portfolio" chart and leave out holdings until digital assets is added (assuming that is the aim?). It feels a bit strange from a UX side of things with just the one selectable dropdown, people may not really understand why it's there. I also feel that "Portfolio" is a better term for a user's collection of IOTAs, colored coins, and NFTs but that's just two cents. Thanks for the changes ~

nicole-obrien:
@maxwellmattryan 
> Hey @nicole-obrien, nice change for the wallet charts. I noticed that whenever the IOTA denomination is selected the y-axis steps display with numbers and units whereas the others display with just numbers. I personally like the numbers with the units as it's more clear, might be worth changing the other charts? Alternatively if you want it without the numbers, we'd have to indicate to users what iota denomination it's in... dropdown or `formatUnitBestMatch()` might be helpful.

Numbers and Units are definitely better, but we don't always have a unit for the currency available, sometimes just the code. I think it's best to leave the other chart for now as that was not part of this issue; we should open another issue to discuss improvements to existing charts.

> I am wondering if it makes sense to put IOTA into the "Portfolio" chart and leave out holdings until digital assets is added (assuming that is the aim?). It feels a bit strange from a UX side of things with just the one selectable dropdown, people may not really understand why it's there. I also feel that "Portfolio" is a better term for a user's collection of IOTAs, colored coins, and NFTs but that's just two cents. Thanks for the changes ~

Had a discussion with UX team and presented several options to them, this is the one we ultimately decided on. The drop down was added for future proofing, as we expect to add more options here. Although, when we do add more assets we need to think about the UX/UI carefully, and have a workshop to flesh out all the details. 

I agree there is multiple names that fit, ideally we want one name to represent the balance (amount) and one to represent value (converted). I think we can come back to the naming at the same time as the additional assets and flesh that out too.

maxwellmattryan:
@nicole-obrien 

> Numbers and Units are definitely better, but we don't always have a unit for the currency available, sometimes just the code. I think it's best to leave the other chart for now as that was not part of this issue; we should open another issue to discuss improvements to existing charts.

Sure, that makes more sense. Haven't looked at the code yet, but does it convert larger amounts into a best match or does it still use `Mi`? The `BTC` chart has quite a few decimals, although I think shorter is probably more readable if it's manageable. 

> Had a discussion with UX team and presented several options to them, this is the one we ultimately decided on. The drop down was added for future proofing, as we expect to add more options here. Although, when we do add more assets we need to think about the UX/UI carefully, and have a workshop to flesh out all the details.
>
> I agree there is multiple names that fit, ideally we want one name to represent the balance (amount) and one to represent value (converted). I think we can come back to the naming at the same time as the additional assets and flesh that out too.

That sounds good to me! 👍

nicole-obrien:
@maxwellmattryan 
> Sure, that makes more sense. Haven't looked at the code yet, but does it convert larger amounts into a best match or does it still use `Mi`? The `BTC` chart has quite a few decimals, although I think shorter is probably more readable if it's manageable.

Yeah, uses the best match.

maxwellmattryan:
Sweet! Everything looks good to me, so should be able to merge once conflicts are settled.

maxwellmattryan (approved):
Great job! 👍

PR #1492, opened 2021-09-23
feat: add deep links

cvarley100:
# Description of change

Adds deep linking to Firefly. Users can fill send data by clicking a link prefixed with `iota://` from outside Firefly.

This PR introduces a new format where the `host` part of the url is now a scope for the operation (so in the future we could also support `iota://did/ ` `iota://chat/` etc

The operation being performed is supplied as the first part of the path after the host section.

e.g. `iota://wallet/send/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1000`

In addition there is an optional `unit` parameter, if not supplied the `amt` is assumed to be `i` units

e.g. `iota://wallet/send/atoi1qzwderym4mesvs7835l52vk7jl7tz264j5yphny8kmeg6xn3vz2w5h67dwl?amt=1.5&unit=Mi`

Fixes issues #884 #700 

- Feat

Tested on Mac

## Change checklist


nicole-obrien:
I guess this can be fleshed out when we have an actual specification for the url scheme. But should we make the query parameters full-length? i.e. `amount` instead of `amt`

nicole-obrien:
If I open an IOTA link with one unit, and open another identical link except with a different unit; the unit updates on screen but the fiat value does not update, making me think the underlying tracked amount is still using the old unit.

nicole-obrien:
Might be one for the future: but when the PR for sending fiat amount is merged, will the feature work with those units too? And if so, what if the unit specified is not the currency currently selected for conversion rate?

nicole-obrien:
Currently rounds the amount to the nearest whole number. i.e. in your example 1.5 Mi rounds to 2 Mi. Tested with other amounts and units too.

nicole-obrien:
Deep link persists after 1 log out but not two.  After following link, completing transfer and logging out; when I log into another profile the send form is filled in with the same details as the link.

packages/desktop/electron/main.js
```diff
@@ -439,13 +439,20 @@ app.on('open-url', (event, url) => {
...
-ipcMain.on('deep-link-request', () => {
+ipcMain.on('check-deep-link-request-exists', () => {
     if (deepLinkUrl) {
         windows.main.webContents.send('deep-link-params', deepLinkUrl)
         deepLinkUrl = null
     }
 })
 
+/**
+ * Proxy deep link event to the wallet application
```
cvarley100, line 451: Might need a different comment

packages/shared/routes/dashboard/Dashboard.svelte
```diff
@@ -50,14 +51,15 @@
         }
     })
 
-    accountsLoaded.subscribe(() => {
+    const unsubscribeFromAccountsLoaded = accountsLoaded.subscribe(() => {
```
cvarley100, line 54: Why `unsubscribeFromAccountsLoaded`?

packages/shared/routes/dashboard/Dashboard.svelte
```diff
@@ -156,11 +163,11 @@
                 dashboardRoute.set(tab)
             }
         }
-        if (!$appSettings.deepLinking) {
+        if (mounted && !$appSettings.deepLinking) {
```
cvarley100, line 166: Why is this check for mounting needed? In case accounts load before it mounts?

packages/desktop/electron/preload.js
```diff
@@ -322,13 +322,14 @@ try {
...
             if (!listeners) {
                 listeners = eventListeners[event] = []
-                ipcRenderer.on(event, (e, args) => {
-                    listeners.forEach((call) => {
-                        call(args)
-                    })
-                })
             }
             listeners.push(callback)
+            ipcRenderer.removeAllListeners(event)
```
cvarley100, line 327: 👍

packages/shared/routes/dashboard/Dashboard.svelte
nicole-obrien: I think this is an artifact from when there was multiple resource leaks and I was getting a lot of unexpected behaviour. I will try to think if this is needed now and test if we remove it.

PR #1603, opened 2021-10-13
feat: Sentry error reporting

maxwellmattryan:
## Summary
This PR includes changes implementing a mechanism to reports errors and crashes. This will help us to identify the common errors and their sources within Firefly, allowing us to achieve a higher quality of software more easily.

:information_source: The user does NOT have to send crash reports; they may opt-in to do so.

### Changelog
```
- Adds functionality to send crash reports via Sentry SDKs for Electron and Rust
- Adds new app setting for sending crash reports (both in code and settings JSON file)
- Adds new onboarding page (new installations), login screen popup (existing installations), and an advanced settings section
- Adds new platform API methods updateAppSettings and getMachineId
- Adds new arguments sendCrashReports and machineId to initialization function for actor system
- Adds new IActorHandler and IWalletActor interfaces for clearer typing
- Modifies GitHub workflow to include necessary Sentry information
- Changes platform API object from Platform to Electron in desktop-side code
- Isolates Webpack config module exports for main, renderer, and preload contexts
```

## Relevant Issues
- #2054
- #2055

## Type of Change

### Platforms
- __Desktop__
	- [x] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions
- "Shared diagnostic reports" setting is currently enabled and user has error (should send)
- "Shared diagnostic reports" setting is currently disabled and user has error (should NOT send)
- "Shared diagnostic reports" setting is toggled to enabled and user has error without prior restart (should NOT send)
- "Shared diagnostic reports" setting is toggled to disabled and user has error without prior restart (should send)

## Checklist

### DIFF
--- a/.github/workflows/build-desktop.yml
+++ b/.github/workflows/build-desktop.yml
@@ -76,7 +76,7 @@ jobs:
       run: yarn build:binding
       working-directory: packages/backend/bindings/node
       env:
-        SENTRY_DSN: ${{ secrets.SENTRY_PROD_BACKEND_DSN }}
+        SENTRY_DSN: ${{ secrets.SENTRY_DSN_PROD_BACKEND }}
 
     - name: Bundle desktop JS
       run: yarn build
@@ -84,7 +84,7 @@ jobs:
       env:
         HARDCODE_NODE_ENV: true
         SENTRY: true
-        SENTRY_DSN: ${{ secrets.SENTRY_PROD_DESKTOP_DSN }}
+        SENTRY_DSN: ${{ secrets.SENTRY_DSN_PROD_DESKTOP }}
         SENTRY_ENVIRONMENT: 'prod'
         SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
 


packages/backend/bindings/node/native/src/lib.rs
```diff
@@ -77,15 +77,21 @@ declare_types! {
             let clone_actor_id = actor_id.clone();
             let storage_path = match cx.argument_opt(1) {
                 Some(arg) => {
-                    Some(arg.downcast::<JsString>().or_throw(&mut cx)?.value())
-                }
+                     Some(arg.downcast::<JsString>().or_throw(&mut cx)?.value())
```
rajivshah3, line 80:
```suggestion
                    Some(arg.downcast::<JsString>().or_throw(&mut cx)?.value())
```

packages/backend/src/lib.rs
```diff
@@ -101,11 +104,37 @@ impl TryFrom<&str> for EventType {
...
 pub async fn init<A: Into<String>>(
     actor_id: A,
     storage_path: Option<impl AsRef<Path>>,
+    send_diagnostics: Option<bool>,
     message_receiver: Arc<Mutex<Sender<String>>>,
 ) {
+    let send_diagnostics = match send_diagnostics {
+        Some(arg) => arg,
+        None => false,
+    };
```
rajivshah3, line 129: There's a warning from `cargo clippy` suggesting that this can be written more concisely:

```
warning: this pattern reimplements `Option::unwrap_or`
   --> src/lib.rs:126:28
    |
126 |       let send_diagnostics = match send_diagnostics {
    |  ____________________________^
127 | |         Some(arg) => arg,
128 | |         None => false,
129 | |     };
    | |_____^ help: replace with: `send_diagnostics.unwrap_or(false)`
    |
    = note: `#[warn(clippy::manual_unwrap_or)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_unwrap_or
```

maxwellmattryan: Thank you for the suggestion, Rajiv! I've added it to the PR :slightly_smiling_face:

packages/desktop/electron/main.js
```diff
@@ -38,10 +56,25 @@ let lastError = {}
...
+            Sentry.captureException(
+                new Error(
+                    JSON.stringify({
+                        type: errorType,
+                        message: error.message || error.reason || error,
+                        stack: error.stack || undefined,
+                        diagnostics,
+                    })
+                )
+            )
```
rajivshah3, line 77: I think it might be better if we capture the original error and then add the error type as [context](https://docs.sentry.io/platforms/javascript/guides/electron/enriching-events/context/#passing-context-directly):

```suggestion
            Sentry.captureException(error, { type: errorType })
```

The diagnostics probably don't need to be added as they should be collected automatically by Sentry

packages/desktop/electron/main.js
```diff
@@ -38,10 +56,25 @@ let lastError = {}
...
         lastError = {
-            diagnostics: getDiagnostics(),
-            errorType,
+            diagnostics,
             error,
+            errorType,
+        }
+
+        if (SEND_CRASH_REPORTS) {
+            Sentry.captureException(
```
rajivshah3, line 68: Are we not able to use the `captureException` from `../sentry`?

packages/shared/locales/en.json
```diff
@@ -406,6 +411,12 @@
                 "title": "Error log",
                 "description": "View errors to debug an issue"
             },
+            "crashReporting": {
+                "title": "Crash reports",
+                "body": "Help the developers improve Firefly by automatically sending diagnostic data when an error or crash occurs",
+                "checkbox": "Send crash reports",
+                "advice": "Please restart the Firefly application to {sendCrashReports, select, true {start sharing crash reports} false {stop sharing crash reports} other {apply this change}}"
```
rajivshah3, line 418: Not sure whether this message needs to be clarified — on macOS, closing the window and re-opening Firefly (as opposed to quitting) may not apply this change:

https://github.com/iotaledger/firefly/blob/366a5d503e8671b6db46d2c9eca4314669f1aec1/packages/desktop/electron/main.js#L326-L333

packages/desktop/electron/main.js
rajivshah3: The stacktrace may also include the OS username, so if we capture the original error I think it should be easier for Sentry's data scrubbing to remove

rajivshah3: Actually I might need to investigate this further :thinking: Sentry seems to report both as "non-error" types

maxwellmattryan: I am happy to make this change. Just let me know after your investigation :mag:

packages/desktop/electron/main.js
maxwellmattryan: We can :+1:

packages/shared/locales/en.json
maxwellmattryan: Yes that's true. I suppose it wouldn't really matter to have the text hint for this; wouldn't say users particularly care about it. Will remove!

packages/shared/components/popups/CrashReporting.svelte
```diff
@@ -0,0 +1,34 @@
...
+    let sendCrashReports = true
+
+    const closePopup = () => {
+        isAwareOfCrashReporting.set(true)
+
+        closePopupInternal()
+    }
+
+    const handleConfirmClick = () => {
+        appSettings.set(<AppSettings>{ ...$appSettings, sendCrashReports, })
```
rajivshah3, line 18: If the user checks the checkbox, I think we need to tell the user to restart Firefly (or use [`app.relaunch()`](https://www.electronjs.org/docs/v14-x-y/api/app#apprelaunchoptions), otherwise Sentry won't be initialized for the current session

maxwellmattryan: Yes, this was intentional. Maybe we could simply say something like "This will take effect the next time you start Firefly" rather then telling them to restart? If not that and we add the text hint here for "Please restart Firefly...", then we should also add it back to the settings.

rajivshah3: Yeah, I think we might want to add it back to the settings in this case

rajivshah3 (approved):
Looks good!

PR #1604, opened 2021-10-18
feat: Developer profile

maxwellmattryan:
# Description of change
This PR adds quite a few things, mainly to accommodate functionality for developer profiles, which at the moment really means more advanced network configuration. The main feature here is the ability to switch networks in a single profile, doing two things: resetting your wallet state (i.e. accounts, transaction history, balances) and logging you out. Upon logging in, similarly as the very first time, you will find a single account / wallet with the correct receive address in the network you just switched to (Bech32 HRP, etc.). 

It is possible to make a developer profile with a Ledger device, but it is not currently possible to switch networks. This is because of a Stronghold error, however I don't see this really affecting too many users. Perhaps we can add support at a later time.

## Notes
I want to add a special [`Tuple` type def](https://stackoverflow.com/questions/52489261/typescript-can-i-define-an-n-length-tuple-type) that would be useful for the node authentication (to match `wallet.rs`), but this requires TypeScript 4.1.0+. Right now Firefly and `wallet.rs` have slightly different authentication types although Firefly is transforming the data accordingly in the [underlying logic](https://github.com/iotaledger/firefly/blob/develop/packages/shared/lib/typings/account.ts#L276) of the API call.

## Changes
- Adds ability to create developer profile (via checkbox in profile naming onboarding page)
- Adds `migrateObjects` function to handle when the type definition for any persisted data changes (protects against `ReferenceError`s)
    - ___NOTE___: _Profile data is migrated on login so that any discrepancies in the data structure between Firefly versions don't create errors or corrupt data. I think in the future we should define a lightweight migration / Svelte store versioning system so that more objects and more complex migrations can be done._
- Changes "Node configuration" to "Network configuration" in the settings navigation
- Changes advanced settings page with more network metadata displayed
- Enhances the adding node popup for more comprehensive authentication and configuration
- Adds network switch popup for guiding (dev profile only) users through switching networks and resetting the wallet
- Adds node info popup accessible by clicking "View info" in the node properties menu
    - "View details" was changed to "Edit details" to keep things clear
- Adds network name to network indicator modal (if dev profile) 
- Refactors [`onCreateAccount`](https://github.com/iotaledger/firefly/blob/4ab6c3ce7a38791dca1faa39421709b15d30ae50/packages/shared/routes/dashboard/wallet/Wallet.svelte#L377) for creating individual wallet accounts

None

## Type of change

I __have__ tested switching between the mainnet and devnet in a single developer profile and regular profile, however I __have NOT__ tested any of the functionality with a private node. I suppose it wouldn't be a bad time to really dig into setting a node or two... 🤔 

Tested on:
- MacOS (Catalina 10.15.7)

## Change checklist


packages/shared/lib/wallet.ts
```diff
@@ -586,6 +584,28 @@ export const asyncCreateAccount = (): Promise<void> =>
         )
     })
 
+export const asyncRemoveWalletAccount = (accountId: string): Promise<void> =>
```
maxwellmattryan, line 587: These two API endpoint wrappers were added previously in `fix/profile-deletion` (#1209), but they're needed here for resetting the wallet state when the network switches.

nicole-obrien (changes requested):
I just found an issue: I have switched from devnet to mainnet and the address displayed to receive to is still a devnet address; does not change on refresh.

nicole-obrien:
Seems the above issue is a combination of things happening. First the when switching from network a to b, using a node url in the official list of the network b, while the official node checklist is unticked, it causes the user added node to be removed whenever you log in. 

This is actually the case in general when having the official node selection unticked but manually adding one from the list will work up until the user logs out and back in.

Then because the client options are not complete, wallet.ts prevents the user/app from refreshing the recieve address. I think it would be better to allow this, as long as we know what network they are connected to?

Also noticed that you can have a trailing slash at the end of urls and it adds that url as a duplicate.

Also I think that you should not display the official node list check box on private networks.

nicole-obrien (commented):
Great just going to test this a bit more!

nicole-obrien (approved):
Looks good to go! 
Will be good to get some feedback from the community of how we can improve and what features they want from a dev profile.

maxwellmattryan:
> Looks good to go!
Will be good to get some feedback from the community of how we can improve and what features they want from a dev profile.

That will be really helpful. Thanks @nicole-obrien for the review!

PR #1670, opened 2021-11-30
fix: swallow receiving on a closed channel error

### LINKED ISSUE #1433: ClientError: {}
### Issue description

When logging out of a profile this error is added to the error log, the

It starts from here https://github.com/iotaledger/firefly/blob/e83c3415deab0b5e25c457a652ef7fd428a158a6/packages/backend/bindings/node/native/src/lib.rs#L62  and the error message is `receiving on a closed channel`
Here https://github.com/iotaledger/firefly/blob/e83c3415deab0b5e25c457a652ef7fd428a158a6/packages/shared/lib/shell/walletApi.ts#L139 it gets converted to this `ClientError` and since the message is only a string, JSON.stringify() removes it

### Firefly version

dev

### OS details

Windows

### Can the issue be reliably reproduced?

Yes

1. Log into a profile
2. Logout
3. Open Error log

### Expected behaviour

There should be no error, the polling should probably stop before or maybe we just ignore this particular error?
Apart from that the error message should be visible

### Actual behaviour

Error is visible in the log, but doesn't really help

### Errors

_No response_

cvarley100:
# Description of change

Due to [the way the poll function is used](https://github.com/iotaledger/firefly/blob/9e136e56a5a16bf4f83b2c32718bd0a440ead02d/packages/backend/bindings/node/index.ts#L72) when sending messages to the backend, poll always returns after destroying the actor. There's no real reason to fundamentally refactor the approach here so we can instead swallow the associated error message.

Fixes #1433 

## Type of change

Tested on Mac dev

## Change checklist

### DIFF
--- a/packages/backend/bindings/node/index.ts
+++ b/packages/backend/bindings/node/index.ts
@@ -90,7 +90,7 @@ export function init(id: string, storagePath?: string): { destroy: () => void; r
         runtime,
         (error, data) => {
             // Swallow error thrown when poll returns after the actor is destroyed
-            if (error && error.message.includes("receiving on a closed channel")) return
+            if (error && error.message.includes('receiving on a closed channel')) return
             const message = error || data
             // @ts-ignore
             onMessageListeners.forEach((listener) => listener(message))
--- a/packages/backend/bindings/node/index.ts
+++ b/packages/backend/bindings/node/index.ts
@@ -66,10 +66,10 @@ const onMessageListeners: ((payload: MessageResponse) => void)[] = []
 
 function _poll(
     runtime: typeof addon.ActorSystem,
-    cb: (error: string, data: unknown) => void,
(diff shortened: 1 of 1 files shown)


packages/backend/bindings/node/index.ts
```diff
@@ -89,6 +89,8 @@ export function init(id: string, storagePath?: string): { destroy: () => void; r
     _poll(
         runtime,
         (error, data) => {
+            // Swallow error thrown when poll returns after the actor is destroyed
+            if (error && error.message.includes("receiving on a closed channel")) return
```
rajivshah3, line 93:
```suggestion
            if (error && error.message.includes('receiving on a closed channel')) return
```

`prettier` is complaining about the double quotes

rajivshah3 (approved):

PR #1671, opened 2021-11-30
update(desktop): Get more accurate OS version info on macOS

rajivshah3:
# Description of change

We currently use `os.release()`, but on macOS this gives us the Darwin version, which then needs to be mapped to a major release of macOS. This PR uses `sw_vers -productVersion` to get the exact macOS version, as this may help debug issues in the future. It also allows us to get the macOS version without needing to add another entry to the map when a new one is released.

N/A

## Type of change

Tested on macOS

## Change checklist


nicole-obrien (approved):
Works for me on MacOS - version 12.0 Beta (Monterey)

PR #1682, opened 2021-12-04
chore: QR optimization

maxwellmattryan:
# Description of change
From [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki):
> For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be used, as those permit the use of alphanumeric mode, which is 45% more compact than the normal byte mode.

This PR simply adds a small bit of logic to transform data into its uppercase representation. 

None

## Type of change

1. Scan the generated QR code
2. Notice capitalized letters inside the Bech32 address
3. (opt.) Enter capitalized version inside the [explorer](https://explorer.iota.org/devnet)

Tested on:
- MacOS (Catalina 10.15.7)

## Change checklist

### DIFF
--- a/packages/shared/components/QR.svelte
+++ b/packages/shared/components/QR.svelte
@@ -5,7 +5,7 @@
     export let data
     export let size = 150
 
-    let cells
+    let cells: boolean[][]
 
     const createCells = (): void => {
         try {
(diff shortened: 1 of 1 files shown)


Thoralf-M (changes requested):
The optimization doesn't work. Left is the old version with lower case, right this PR with upper case
image
the blocks of the new QR code should be larger like here using https://www.the-qrcode-generator.com/
image

PR #1720, opened 2021-12-13
fix: various export transaction history fixes

nicole-obrien:
# Description of change

For exporting transaction history:
- Make headers more verbose
- Separate date and time into their own columns and add `(UTC)` to time column header
- Switch minus sign around so that it is correctly on outgoing transactions
- Add extra column to show formatted value with units

## Type of change

Manually tested on MacOS

## Change checklist


packages/shared/lib/transactionHistory.ts
```diff
@@ -15,26 +18,41 @@ export const generateTransactionHistoryCsvFromAccount = (
     headerParams: ITransactionHistoryHeaderParameters
 ): string => {
     const headerParts = []
-    headerParams.id && headerParts.push('id')
+    headerParams.id && headerParts.push('message id')
```
cvarley100, line 21: these could be translated

cvarley100 (changes requested):
When you export (at least on Mac) the file system window reopens after every export. It should close and the modal close too. Also not sure why the file name includes "wallet 1", the wallet is called "gaps".

Screenshot 2021-12-14 at 03 28 07

packages/shared/lib/transactionHistory.ts
nicole-obrien: Will do that

nicole-obrien:
> When you export (at least on Mac) the file system window reopens after every export. It should close and the modal close too. Also not sure why the file name includes "wallet 1", the wallet is called "gaps".

It's because its allowing you to save a file for each account consecutively. This was just for the MVP and I have item on backlog to improve this, i.e. allowing them to select which account to export.

Would also be an improvement to add this to the menu on a specific account too, below where we view address history we could have a button to export this specific account.

As for the name of the file I will double check this..

nicole-obrien:
@cvarley100 

I have moved the export to the AccountActions menu and it is now on a per account basis. I have also localised the field names for the csv file.

It's now ready for a rereview 😊

cvarley100 (approved):

PR #1746, opened 2021-12-15
fix: set primary node on new profile

nicole-obrien:
# Description of change

Issue was that staking wasn't loading when the profile was new, until the user went to the networking settings. This was because when the profile is first created it doesn't set any of the official nodes in the official network config to the primary node. I added some code to select a random one as the primary node.

## Type of change

Tested manually on MacOS by recovering a profile using a 24 word mnemonic.

## Change checklist

### DIFF
--- a/packages/shared/lib/network.ts
+++ b/packages/shared/lib/network.ts
@@ -29,13 +29,19 @@ export const CHRYSALIS_DEVNET_BECH32_HRP = 'atoi'
  *
  * @returns {NetworkConfig}
  */
-export const getOfficialNetworkConfig = (type: NetworkType): NetworkConfig => ({
-    network: getOfficialNetwork(type),
-    nodes: getOfficialNodes(type),
-    automaticNodeSelection: true,
-    includeOfficialNodes: true,
-    localPow: true,
-})
+export const getOfficialNetworkConfig = (type: NetworkType): NetworkConfig => {
+    const officialNodes = getOfficialNodes(type)
+    const randIdx = Math.floor(Math.random() * officialNodes.length)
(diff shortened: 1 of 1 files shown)


packages/shared/lib/network.ts
```diff
@@ -29,13 +29,19 @@ export const CHRYSALIS_DEVNET_BECH32_HRP = 'atoi'
...
-export const getOfficialNetworkConfig = (type: NetworkType): NetworkConfig => ({
-    network: getOfficialNetwork(type),
-    nodes: getOfficialNodes(type),
-    automaticNodeSelection: true,
-    includeOfficialNodes: true,
-    localPow: true,
-})
+export const getOfficialNetworkConfig = (type: NetworkType): NetworkConfig => {
+    const officialNodes = getOfficialNodes(type)
+    const randIdx = Math.floor(Math.random() * officialNodes.length)
```
maxwellmattryan, line 34: There is a [`pick`](https://github.com/iotaledger/firefly/blob/feat/staking/packages/shared/lib/utils.ts#L351) function in `lib/utils.ts` that will randomly select the element of an array - we can de-dup some code a bit and substitute it here.

nicole-obrien: @maxwellmattryan that returns the specific element of the array, where I want to modify a specific element in an array and keep the rest of the array too.

Although there is this exact code further down that I can extract into its own function.

nicole-obrien: Made it a little more DRY 😊

maxwellmattryan (approved):
Very noice 👍

PR #1753, opened 2021-12-15
fix: minimum reward time

maxwellmattryan:
# Description of change
- Adds safe object property access with optional chaining (fixed `ReferenceError` bug with `Cannot read property index of undefined`)
- Adds tooltip in `StakingManager` if account will not reach min rewards for either airdrop (with calculated amount of IOTAs still needed to stake to reach min)
- Changes tooltip text for `StakingIndicator` if events are ended and no account has reached minimum
- Fixes missing store variable reset statements inside `resetParticipation()`
- Cleans up some empty `import` statements

None

## Type of change

Tested on: 
- MacOS (10.15.7)

## Change checklist

### DIFF
--- a/packages/shared/components/AccountTile.svelte
+++ b/packages/shared/components/AccountTile.svelte
@@ -1,9 +1,9 @@
 <script lang="typescript">
-    import { get } from 'svelte/store'
     import { Icon, Text, Tooltip } from 'shared/components'
     import { localize } from 'shared/lib/i18n'
     import {
         formatStakingAirdropReward,
+        getAccountParticipationAbility,
         getMinimumAirdropRewardInfo,
         getStakingEventFromAirdrop,
         getTimeUntilMinimumAirdropReward,
@@ -18,12 +18,16 @@
         stakedAccounts,
         stakingEventState,
     } from 'shared/lib/participation/stores'
-    import { ParticipationEventState, ParticipationOverview, StakingAirdrop } from 'shared/lib/participation/types'
+    import {
+        AccountParticipationAbility,
+        ParticipationEventState,
+        ParticipationOverview,
+        StakingAirdrop,
+    } from 'shared/lib/participation/types'
     import { openPopup } from 'shared/lib/popup'
     import { getBestTimeDuration } from 'shared/lib/time'
     import { formatUnitBestMatch } from 'shared/lib/units'
     import { capitalize } from 'shared/lib/utils'
-    import { wallet } from 'shared/lib/wallet'
     import type { WalletAccount } from 'shared/lib/typings/wallet'
 
     export let name = ''
@@ -62,7 +66,7 @@
          * accounts on the wallet Svelte store.
          */
         const account = _getAccount($stakedAccounts)
-        if (account) {
+        if (account && getAccountParticipationAbility(account) !== AccountParticipationAbility.NoHasDustAmount) {
             const accountOverview = $participationOverview.find((apo) => apo.accountIndex === account?.index)
             isBelowMinimumStakingRewards =
                 accountOverview?.assemblyRewardsBelowMinimum > 0 || accountOverview?.shimmerRewardsBelowMinimum > 0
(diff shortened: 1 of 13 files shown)


packages/shared/components/popups/StakingManager.svelte
```diff
@@ -240,6 +257,14 @@
                         <div class="bg-green-100 rounded-2xl">
                             <Icon icon="success-check" width="19" height="19" classes="text-white" />
                         </div>
+                    {:else if getAccountParticipationAbility(account) === AccountParticipationAbility.NoWillNotReachMinAirdrop}
```
nicole-obrien, line 260: This statement will always resolve false because the condition is not included in the parent if statement.

nicole-obrien (changes requested):

packages/shared/lib/participation/staking.ts
```diff
@@ -370,7 +379,107 @@ export const getTimeUntilMinimumAirdropReward = (account: WalletAccount, format:
...
+ *
+ * @returns {boolean}
+ */
+export const canAccountReachMinimumAirdrop = (account: WalletAccount): boolean => {
+    if (!account) return false
+
+    const [minRewards, minAirdrop, amountStaked] = getMinimumAirdropRewardInfo(account)
+    if (!minAirdrop) return true
+
+    const timeRequired = calculateTimeUntilMinimumReward(minRewards, minAirdrop, amountStaked)
```
nicole-obrien, line 449: There is an existing function for this logic in the same file called: getTimeUntilMinimumAirdropReward

packages/shared/components/AccountTile.svelte
```diff
@@ -135,7 +150,6 @@
                 const timeNeeded = <number>getTimeUntilMinimumAirdropReward(account)
                 const remainingTime =
                     airdrop === StakingAirdrop.Assembly ? $assemblyStakingRemainingTime : $shimmerStakingRemainingTime
-
                 if (timeNeeded > remainingTime) {
```
nicole-obrien, line 153: why not use the function canAccountReachMinimumAirdrop from staking.ts

nicole-obrien (commented):
Some more comments

packages/shared/lib/participation/staking.ts
```diff
@@ -370,7 +379,107 @@ export const getTimeUntilMinimumAirdropReward = (account: WalletAccount, format:
     if (!account) return format ? getBestTimeDuration(0) : 0
 
     const [minRewards, minAirdrop, amountStaked] = getMinimumAirdropRewardInfo(account)
```
nicole-obrien, line 381: Using the current implementation of getMinimumAirdropRewardInfo, this will return [0, undefined, 0] which will intern cause this function to return 0. This is because you inside of getMinimumAirdropRewardInfo you only get the information if the account has or is participating in an airdrop.

packages/shared/lib/participation/staking.ts
```diff
@@ -370,7 +379,107 @@ export const getTimeUntilMinimumAirdropReward = (account: WalletAccount, format:
...
+ * @method canAccountReachMinimumAirdrop
+ *
+ * @param {WalletAccount} account
+ *
+ * @returns {boolean}
+ */
+export const canAccountReachMinimumAirdrop = (account: WalletAccount): boolean => {
+    if (!account) return false
+
+    const [minRewards, minAirdrop, amountStaked] = getMinimumAirdropRewardInfo(account)
```
nicole-obrien, line 446: Using the current implementation of getMinimumAirdropRewardInfo, this will return [0, undefined, 0] which will intern cause this function to return 0. This is because you inside of getMinimumAirdropRewardInfo you only get the information if the account has or is participating in an airdrop.

This will then lead this function to return true when calculating if an account can reach the minimum airdrop before they ever stake it. Making the warning tooltip on the ManageStaking never show. (well once you fix my previous comment)

nicole-obrien (changes requested):
Some more blockers.

packages/shared/components/AccountTile.svelte
maxwellmattryan: Because the data was needed later for the locale, however this could definitely be re-written.

packages/shared/lib/participation/staking.ts
maxwellmattryan: Yes, the intention was for `calculateTimeUntilMinimumReward` to make the calculation given prepared data whereas `getTimeUntilMinimumReward` is an exposed function that does a bit more logic to prepare the data needed then internally calls `calculateTimeUntilMinimumReward`.

cvarley100 (approved):

PR #1775, opened 2021-12-17
fix: use account ID instead of index when getting migration address for ledger

laumair:
# Description of change

Generating migration address for ledger was erroring out. This PR fixes the issue. 

N/A

## Type of change

- Tested manually by generating migration address. 
- Did not test generating migration address during the bundle mining & bundle creation process (but it still is passing id so it should work fine)

## Change checklist

### DIFF
--- a/packages/shared/lib/typings/migration.ts
+++ b/packages/shared/lib/typings/migration.ts
@@ -224,7 +224,7 @@ export function getMigrationAddress(
     bridge: Bridge,
     __ids: CommunicationIds,
     ledgerPrompt: boolean,
-    accountIndex: string | number
+    accountIdentifier: string | number
 ): Promise<string> {
     return bridge({
         actorId: __ids.actorId,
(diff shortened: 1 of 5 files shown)


packages/shared/lib/typings/migration.ts
```diff
@@ -224,7 +224,7 @@ export function getMigrationAddress(
     bridge: Bridge,
     __ids: CommunicationIds,
     ledgerPrompt: boolean,
-    accountIndex: number
+    accountIndex: string | number
```
maxwellmattryan, line 227: Should we rename this parameter to `accountId` rather than `accountIndex`?

laumair: Yes, good catch @maxwellmattryan.

laumair: Fixed in f3b9dad

maxwellmattryan: Nice. I should've mentioned that there's also this [`AccountIdentifier`](https://github.com/iotaledger/firefly/blob/develop/packages/shared/lib/typings/account.ts#L43) type in `typings/account.ts` we can use here!

maxwellmattryan (approved):
Code LGTM!

PR #1785, opened 2021-12-18
fix(shared): Use ICU message syntax for staking duration messages

rajivshah3:
# Description of change

Languages such as Polish and Ukrainian have 4 [plural categories](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html) (one, few, many, other) instead of 2 (one, other) in languages like English. The current logic in `getBestTimeDuration` assumes languages have 2 plural categories, which causes strings to show with incorrect grammar in languages that have more than 2 plural categories. 

This PR uses ICU message syntax (similar to #64) to handle plurals for the staking duration messages and refactors `getBestTimeDuration` accordingly.

Mentioned by Beniz in Discord

## Type of change

Tested on macOS

## Change checklist

### DIFF
--- a/packages/shared/lib/time.ts
+++ b/packages/shared/lib/time.ts
@@ -30,24 +30,28 @@ export const getBestTimeDuration = (
     millis: number,
     noDurationUnit: 'days' | 'hours' | 'minutes' | 'seconds' = 'days'
 ): string => {
-    const noDurationInLocale = localize(`general.time.${noDurationUnit}`)
-    if (Number.isNaN(millis)) return `0 ${noDurationInLocale}`
+    const noDurationInLocale = localize(`general.time.${noDurationUnit}`, { values: { [noDurationUnit]: '0' } })
+    if (Number.isNaN(millis)) return noDurationInLocale
 
     const inDays = millis / (HOURS_PER_DAY * MINUTES_PER_HOUR * SECONDS_PER_MINUTE * MILLISECONDS_PER_SECOND)
-    if (inDays > 1) return `${Math.ceil(inDays)} ${localize('general.time.days')}`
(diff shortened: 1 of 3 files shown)


packages/shared/locales/en.json
```diff
@@ -1000,16 +1000,11 @@
         "version": "Version {version}",
         "unknown": "Unknown",
         "none": "None",
-        "days": "day(s)",
```
rajivshah3, line 1003: This string appears to be unused and also assumes only 2 plural categories, so I removed it

packages/shared/lib/time.ts
```diff
@@ -30,24 +30,28 @@ export const getBestTimeDuration = (
     millis: number,
     noDurationUnit: 'days' | 'hours' | 'minutes' | 'seconds' = 'days'
 ): string => {
-    const noDurationInLocale = localize(`general.time.${noDurationUnit}`)
-    if (Number.isNaN(millis)) return `0 ${noDurationInLocale}`
```
rajivshah3, line 34: Changed this because we shouldn't assume spacing (for example, Chinese does not separate numbers and characters with a space, so `1 day` is `1天)

maxwellmattryan: Yes, you're right. We will need to change some logic in the code for displaying the time estimates appropriately. Good catch, @rajivshah3 👍

maxwellmattryan: @rajivshah3 I have added a fix for this in `feat/staking` that does not rely on or use this whitespace anymore. There is an entry on Crowdin that suggests another way of writing the `en.json` locale file, so there may be some conflicts.

cvarley100:
Included in @maxwellmattryan's PR

PR #1813, opened 2021-12-20
feat: keep track of the last logged in profile

jeeanribeiro:
# Description of change

Keep track of the last logged-in profile using persistent storage of the active profile id

Resolves #1562
Closes https://github.com/iotaledger/firefly/issues/2050

## Type of change

- Update (a change that updates existing functionality)

- Select a profile
- Refresh the application
- Enter pin screen is the first loaded one and the select profile is the first step one
- Click on the previous button
- Refresh the application
- Select profile screen is the first loaded one

## Change checklist

### DIFF
--- a/packages/shared/routes/login/Login.svelte
+++ b/packages/shared/routes/login/Login.svelte
@@ -3,7 +3,8 @@
     import { Transition } from 'shared/components'
     import { SelectProfile, EnterPin } from './views/'
     import { Locale } from 'shared/lib/typings/i18n'
-    import { migrateProfile } from 'shared/lib/profile'
+    import { migrateProfile, activeProfileId } from 'shared/lib/profile'
+    import { get } from 'svelte/store'
 
     export let locale: Locale
 
@@ -16,7 +17,7 @@
 
     const dispatch = createEventDispatcher()
 
-    let state: LoginState = LoginState.Init
+    let state: LoginState = get(activeProfileId) ? LoginState.EnterPin : LoginState.Init
     let stateHistory = []
 
     const _next = (event) => {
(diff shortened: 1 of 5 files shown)


packages/shared/routes/login/Login.svelte
```diff
@@ -49,6 +50,7 @@
         if (prevState) {
             state = prevState
         } else {
+            state = LoginState.Init
```
begonaalvarezd, line 53: why is this line needed here?

begonaalvarezd (commented):
Was thinking that it might be a good idea not clearing up the active profile on logout? The app triggers logout in 2 scenarios: 
- User explicitly clicks on logout
- Auto logout after timeout

I think in both scenarios we have also suggest the previous logged profile? 
wdyt?

packages/shared/routes/login/Login.svelte
jeeanribeiro: If you start the app it does not have a previous state, so when the EnterPin screen is active I need a way to return to the main screen with the profile list

packages/shared/routes/login/Login.svelte
```diff
@@ -19,9 +20,19 @@
     let state: LoginState = LoginState.Init
     let stateHistory = []
 
-    const _next = (event) => {
+    onMount(() => {
+        if (get(activeProfileId)) {
```
begonaalvarezd, line 24: Just realized we have to check if the `activeProfileId` exists, people can delete profiles directly removing the folder in their system

begonaalvarezd: resolved

begonaalvarezd (approved):

PR #1857, opened 2021-12-22
feat: transaction search and filtering

jeeanribeiro:
# Description of change

Adds transaction counter, search (transaction addresses and balance), and filtering (all, incoming, outgoing)

Resolves https://github.com/iotaledger/firefly/discussions/655
Closes https://github.com/iotaledger/firefly/issues/2053

## Type of change

- Have some transaction history in the account
- Access account history screen
- Test filters (all, incoming and outgoing)
- Test search by balance or transaction addresses
- Test search + filter

## Change checklist


packages/shared/components/icon/icons.js
```diff
@@ -1174,4 +1174,15 @@ export const icons = {
             },
         ],
     },
+    search: {
+        width: 20,
+        height: 20,
+        path: [
+            {
+                d: 'M11.1922 12.6064C10.0236 13.4816 8.57236 14 7 14C3.13401 14 0 10.866 0 7C0 3.13401 3.13401 0 7 0C10.866 0 14 3.13401 14 7C14 8.57232 13.4816 10.0236 12.6064 11.1921L16.0001 14.5858C16.3906 14.9763 16.3906 15.6095 16.0001 16C15.6096 16.3905 14.9764 16.3905 14.5859 16L11.1922 12.6064ZM12 7C12 9.76142 9.76142 12 7 12C4.23858 12 2 9.76142 2 7C2 4.23858 4.23858 2 7 2C9.76142 2 12 4.23858 12 7Z',
```
begonaalvarezd, line 1182: Im going to update the path of this icon to follow the rules of the other icons in the UI Kit

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -92,14 +124,31 @@
                     </button>
                 {/if}
             </div>
+            <div class="flex flex-row justify-between text-white mt-4">
+                <ul class="flex flex-row justify-between">
+                    {#each filters as filter, i}
+                        <li on:click={() => activeFilterIndex = i}>
+                            <Text type="p" overrideColor classes="cursor-pointer mr-8 capitalize
+                            {activeFilterIndex === i ? "text-blue-500 border-b-2 border-blue-500 border-solid" : "text-gray-500 hover:text-gray-600"}">
+                                {filter}
```
begonaalvarezd, line 133: if we use `{filter}` we cant translate it. Better localize it

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -72,6 +72,38 @@
...
+    let searchActive = false
+    $: searchValue = searchActive ? searchValue : ''
+
+    let filteredTransactions = transactions
+    $: switch (activeFilterIndex) {
+        case 0:
+            filteredTransactions = transactions
+            break
+        case 1:
+            filteredTransactions = transactions.filter(transaction => transaction.payload.data.essence.data.incoming)
```
begonaalvarezd, line 88: to avoid possible crashes might be good using `transaction?.payload?.data?.essence?.data?.incoming` everywhere

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
begonaalvarezd: Updated

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -72,6 +72,38 @@
...
+        default:
+            filteredTransactions = transactions
+    }
+
+    let queryTransactions = filteredTransactions
+    $: if (searchValue) {
+        queryTransactions = filteredTransactions.filter(transaction => {
+            return transaction?.payload.data?.essence?.data?.value?.toString()?.includes(searchValue) ||
+                sendAddressFromTransactionPayload(transaction?.payload).includes(searchValue) ||
+                receiverAddressesFromTransactionPayload(transaction?.payload).find(addr => addr.includes(searchValue))
```
begonaalvarezd, line 102: What about adding here also a match for `messageID`?

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -72,6 +80,41 @@
...
+
+    let queryTransactions = filteredTransactions
+    $: if (searchValue) {
+        queryTransactions = filteredTransactions.filter(transaction => {
+            if (transaction?.payload?.type === 'Transaction') {
+                return transaction?.payload?.data?.essence?.data?.value?.toString()?.includes(searchValue) ||
+                    sendAddressFromTransactionPayload(transaction?.payload).includes(searchValue) ||
+                    receiverAddressesFromTransactionPayload(transaction?.payload).find(addr => addr.includes(searchValue))
+            }
+            return transaction?.id?.includes(searchValue)
```
begonaalvarezd, line 113: This will only search the messageID if the transaction is not of type `Transaction`

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -72,6 +80,41 @@
...
+            break
+        case 2:
+            filteredTransactions = transactions.filter(transaction => !getIncomingFlag(transaction?.payload))
+            break
+        default:
+            filteredTransactions = transactions
+    }
+
+    let queryTransactions = filteredTransactions
+    $: if (searchValue) {
```
begonaalvarezd, line 106: perhaps we could add a `debounce`?

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -92,14 +135,37 @@
...
+                        </li>
+                    {/each}
+                </ul>
+                <button on:click={() => (searchActive = !searchActive)}>
+                    <Icon icon="search" classes="text-gray-500 hover:text-gray-600 dark:text-white dark:hover:text-gray-100
+                    cursor-pointer ml-2" />
+                </button>
+                <div class="z-0 flex items-center absolute left-0 transition-all {searchActive ? 'w-full' : 'w-0'} overflow-hidden">
+                    <Icon icon="search" classes="z-10 absolute left-2 text-gray-500" />
+                    <Input bind:value={searchValue} classes="z-0" style="padding: 0.75rem  2.5rem;" />
```
begonaalvarezd, line 155: I would also add a focus when the search opens

nicole-obrien (changes requested):
Currently crashes after the following steps:
- Click wallet
- Click search icon
- Click on any transaction to view details

This is the error in the console:
Screenshot 2022-01-26 at 15 13 03

nicole-obrien (changes requested):
Two more points from me:
- When filtering and then clicking the search bar the current filter is not displayed, but transactions displayed are only the filtered ones. This might be confusing to the user. Maybe we can revisit this at a later time.
- When searching by value it is only checking the raw iota value, but we display to the user a simplified balance e.g. 1343902990 would be 1.30 Gi. Therefore I think the user should also receive results for the balance that is displayed in the transaction history view. so searching for 1.3 would show that result or even searching Gi would show that result to.

jeeanribeiro:
> Two more points from me:
> 
> * When filtering and then clicking the search bar the current filter is not displayed, but transactions displayed are only the filtered ones. This might be confusing to the user. Maybe we can revisit this at a later time.
> * When searching by value it is only checking the raw iota value, but we display to the user a simplified balance e.g. 1343902990 would be 1.30 Gi. Therefore I think the user should also receive results for the balance that is displayed in the transaction history view. so searching for 1.3 would show that result or even searching Gi would show that result to.

Latest changes:
When the search is active the filter is reset.
I've changed the search function to [trinity search algorithm](https://github.com/iotaledger/trinity-wallet/blob/c6d8e798ad503b8592a4e0c1b2955dc253c0e9da/src/shared/libs/iota/transfers.js#L1111)

nicole-obrien (changes requested):
New changes don't seem to be working properly, I would revert back to the previous search function and add condition like this: 
`formatUnitBestMatch(transaction?.payload?.data?.essence?.data?.value).toString()?.includes(searchValue)`

nicole-obrien (approved):
LGTM, thanks for all the changes as requested 🙏

JannemanDev:
I dont know if intentional but search on address only matches when there is a full match not if found as substring.
For example when I have a tx in my history from this address: `atoi1qptd4dlt2870zsa9fn5t98mnt2zsstlert0wd20uesk0h4thzavmgrld8kk`
and I search on `atoi1qptd4dlt2870zsa9fn5t98mnt2zsstlert0wd20uesk0h4thzavmgrld8` (I removed last two chars) it will not show up.
Tested with FireFly 1.4 Alpha 1.

Personally I think it can be handy to search on substring... when you look at an address you only have to remember a few chars (instead of having to copy and paste which takes more time *and* sometimes isn't even possible when its on a different device) most of the time to be able to find it in history.

nicole-obrien:
> I dont know if intentional @nicole-obrien  but search on address only matches when there is a full match not if found as substring.
> 
> For example when I have a tx in my history from this address: `atoi1qptd4dlt2870zsa9fn5t98mnt2zsstlert0wd20uesk0h4thzavmgrld8kk`
> 
> and I search on `atoi1qptd4dlt2870zsa9fn5t98mnt2zsstlert0wd20uesk0h4thzavmgrld8` (I removed last two chars) it will not show up.
> 
> Tested with FireFly 1.4 Alpha 1.
> 
> 
> 
> Personally I think it can be handy to search on substring... when you look at an address you only have to remember a few chars (instead of having to copy and paste which takes more time *and* sometimes isn't even possible when its on a different device) most of the time to be able to find it in history.

That's a valid point, I can see if this is a quick fix.

PR #1900, opened 2021-12-27
fix: pass correct account id during legacy ledger migration

laumair:
# Description of change

This PR fixes the issue where an exception was thrown during legacy ledger migration.

N/A

## Type of change

Manually tested with ledger device

## Change checklist

### DIFF
--- a/packages/backend/Cargo.toml
+++ b/packages/backend/Cargo.toml
@@ -8,7 +8,7 @@ exclude = ["/bindings", "/api-wrapper"]
 [dependencies]
 tokio = { version = "1.12.0", default-features = false }
 once_cell = { version = "1.8.0", default-features = false }
-iota-wallet = { git = "https://github.com/iotaledger/wallet.rs", rev = "79163acbecbf2aeecb8caa99768f4d1437644bed", default-features = false, features = ["stronghold", "ledger-nano", "ledger-nano-simulator", "participation"] }
+iota-wallet = { git = "https://github.com/iotaledger/wallet.rs", rev = "fc6451608db70d9c25c26ee86b67f72dfac83a40", default-features = false, features = ["stronghold", "ledger-nano", "ledger-nano-simulator", "participation"] }
 serde = { version = "1.0.130", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.68", default-features = false }
 riker = "0.4.2"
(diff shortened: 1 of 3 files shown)


packages/backend/Cargo.toml
```diff
@@ -8,7 +8,7 @@ exclude = ["/bindings", "/api-wrapper"]
 [dependencies]
 tokio = { version = "1.12.0", default-features = false }
 once_cell = { version = "1.8.0", default-features = false }
-iota-wallet = { git = "https://github.com/iotaledger/wallet.rs", rev = "054c36fd3281119b22c09da25e175a3d08465873", default-features = false, features = ["stronghold", "ledger-nano", "ledger-nano-simulator", "participation"] }
+iota-wallet = { git = "https://github.com/iotaledger/wallet.rs", rev = "79163acbecbf2aeecb8caa99768f4d1437644bed", default-features = false, features = ["stronghold", "ledger-nano", "ledger-nano-simulator", "participation"] }
```
maxwellmattryan, line 11: We need to be careful when merging this and #1918 (it also changes the wallet.rs commit hash we use).

packages/shared/lib/migration.ts
```diff
@@ -360,16 +360,23 @@ export const findMigrationBundle = (bundleIndex: number): Bundle => {
...
-        const accountId = get(accounts).find((account) => account.index === get(activeProfile).ledgerMigrationCount).id
-
-        api.getMigrationAddress(false, accountId, {
-            onSuccess(response) {
-                resolve(response.payload)
+        api.getAccounts({
+            onSuccess(getAccountsResponse) {
+                api.getMigrationAddress(
+                    false,
+                    getAccountsResponse.payload[get(activeProfile).ledgerMigrationCount].id,
```
maxwellmattryan, line 367: Adding optional chaining here might prevent a runtime exception (just in case for whatever reason the `activeProfile` is not found), like:
```suggestion
                    getAccountsResponse.payload[get(activeProfile)?.ledgerMigrationCount]?.id,
```

cvarley100: What would happen if it is undefined? Error from rust side? We don't want to accidentally generate a migration address on the wrong index

maxwellmattryan: I'd imagine just a rust-side error would be handed back, `Invalid type of action received.` or something like that

cvarley100: I’ll simulate it

cvarley100 (approved):

PR #1978, opened 2021-12-30
fix: ensure syncing message only shows for valid pending transactions

cvarley100:
# Description of change

Block staking on a wallet only if the pending transactions can actually confirm (by comparing UTXO id in message against the UTXO on the address)

## Type of change

Tested invalid pending transactions by double spending (to create one that will never confirm).
Tested valid pending transactions with a normal spend.

## Change checklist

### DIFF
--- a/packages/shared/lib/wallet.ts
+++ b/packages/shared/lib/wallet.ts
@@ -2075,10 +2075,11 @@ export const hasValidPendingTransactions = (account: WalletAccount): boolean => 
         if (msg.payload?.type === 'Transaction') {
             return msg.payload?.data?.essence?.data?.inputs
         }
+        return []
     })
     const unspentOutputs = account?.addresses.filter((a) => a.balance > 0).flatMap((a) => Object.values(a.outputs))
 
     return pendingInputs.some((i) =>
-        unspentOutputs.map((o) => o.transactionId).includes(i.data?.metadata?.transactionId)
+        unspentOutputs.some((o) => o.transactionId === i.data?.metadata?.transactionId)
     )
 }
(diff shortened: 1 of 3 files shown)


packages/shared/lib/wallet.ts
```diff
@@ -2060,3 +2058,27 @@ export const hasPendingTransactions = (account: WalletAccount): boolean => {
...
+    const pendingMessages = account?.messages.filter((m) => !m.confirmed)
+    const pendingInputs = pendingMessages.flatMap((msg) => {
+        if (msg.payload?.type === 'Transaction') {
+            return msg.payload?.data?.essence?.data?.inputs
+        }
+    })
+    const unspentOutputs = account?.addresses.filter((a) => a.balance > 0).flatMap((a) => Object.values(a.outputs))
+
+    return pendingInputs.some((i) =>
+        unspentOutputs.map((o) => o.transactionId).includes(i.data?.metadata?.transactionId)
```
maxwellmattryan, line 2082: I think `unspentOutputs.some((o) => o.transactionId === i.data?.metadata?.transactionId)` might be slightly more efficient.

packages/shared/lib/wallet.ts
```diff
@@ -2060,3 +2058,27 @@ export const hasPendingTransactions = (account: WalletAccount): boolean => {
...
+ *
+ * @returns {boolean}
+ */
+export const hasValidPendingTransactions = (account: WalletAccount): boolean => {
+    if (!account) return false
+    const pendingMessages = account?.messages.filter((m) => !m.confirmed)
+    const pendingInputs = pendingMessages.flatMap((msg) => {
+        if (msg.payload?.type === 'Transaction') {
+            return msg.payload?.data?.essence?.data?.inputs
+        }
```
laumair, line 2077: It should have a default return as well outside `if` because if the payload type is not `Transaction` then it will return `undefined` values.

maxwellmattryan (approved):
LGTM!

PR #2116, opened 2022-01-12
update: Sentry reporting improvements

rajivshah3:


packages/desktop/.sentrycliignore
```diff
@@ -1,4 +1,3 @@
-public/
```
rajivshah3, line 1: I don't think we want to ignore `public/` since that's where our build outputs are located

packages/desktop/webpack.config.js
```diff
@@ -145,12 +146,11 @@ const sentryPlugins = [
     new SentryWebpackPlugin({
         authToken: process.env.SENTRY_AUTH_TOKEN,
         include: '.',
-        release: `firefly-${version}`,
+        release: `Firefly@${version}`,
         ignoreFile: '.sentrycliignore',
-        ignore: ['node_modules', 'webpack.config.js'],
```
rajivshah3, line 150: I think this would conflict with `ignoreFile`

packages/desktop/webpack.config.js
```diff
@@ -145,12 +146,11 @@ const sentryPlugins = [
     new SentryWebpackPlugin({
         authToken: process.env.SENTRY_AUTH_TOKEN,
         include: '.',
-        release: `firefly-${version}`,
+        release: `Firefly@${version}`,
```
rajivshah3, line 149: Sentry recommends using `@` (https://docs.sentry.io/product/cli/releases/#creating-releases)

packages/desktop/webpack.config.js
```diff
@@ -168,7 +168,7 @@ module.exports = [
         },
         mode,
         plugins: [...rendererPlugins, ...sentryPlugins],
-        devtool: prod ? false : 'cheap-module-source-map',
+        devtool: SENTRY && prod ? 'source-map' : 'cheap-module-source-map',
```
rajivshah3, line 171: We want detailed sourcemaps when we submit to Sentry, but we don't really need extensive detail when we're doing local development

maxwellmattryan (approved):
Ready to merge!

PR #2245, opened 2022-01-27
enhancement: setting hints

maxwellmattryan:
# Description of change
While there is only one text hint in the settings, there is soon to be another with the Sentry bug reporting changes. Therefore it is best we create a single re-usable component for this, also slightly improving the style at the same time (the hint is given same attention level as other secondary settings text rather than popping out at the user).

 __Before__
before

__After__
after

_None_

## Type of change

Platforms:
- Ubuntu 21.10

## Change checklist


packages/shared/components/TextHint.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="typescript">
+    import { Icon, Text } from 'shared/components'
+
+    export let classes: string = ''
+    export let icon: 'info' | 'exclamation' | 'none' = 'none'
+    export let hint: string = ''
+</script>
+
+{#if hint !== ''}
```
nicole-obrien, line 9: empty strings are falsey so could just write:
```suggestion
{#if hint}
```

nicole-obrien (approved):
LGTM!

PR #2449, opened 2022-02-28
Revert "chore: update capacitor bindings script"

kraftjs:
Reverts iotaledger/firefly#2424 because it includes a breaking change for mobile.


maxwellmattryan (approved):
Thanks :pray:

Issue #2397, opened 2022-02-21
[Task]: Add workaround to fix Sentry on mobile


amadeu2:
### Task description

Since Sentry isn't implemented on mobile, we must to fix it to be able create account.

### Requirements

- [ ] Create account and use the app.

### Acceptance criteria

TBD

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #2461, opened 2022-03-02
Fix: Mobile Sentry

amadeu2:
## Summary
Currently Sentry is not implemented on Mobile, added temporally optional values to AppSettings store.
This PR solves Android crashes at the moment of create the password.

## Relevant Issues
Closes https://github.com/iotaledger/firefly/issues/2397

## Type of Change

### Platforms
- __Desktop__
	- [x] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [x] iOS
	- [x] Android

### Instructions

## Checklist

### DIFF
--- a/packages/shared/routes/setup/Profile.svelte
+++ b/packages/shared/routes/setup/Profile.svelte
@@ -68,7 +68,7 @@
 
                 const path = await getProfileDataPath($newProfile.name)
                 const machineId = await Platform.getMachineId()
-                const { sendCrashReports } = $initAppSettings ?? {}
+                const { sendCrashReports } = $initAppSettings ?? { sendCrashReports: false }
                 initialise($newProfile.id, path, sendCrashReports, machineId)
                 initialiseMigrationListeners()
             }
--- a/packages/shared/routes/setup/Profile.svelte
+++ b/packages/shared/routes/setup/Profile.svelte
@@ -68,7 +68,7 @@
 
                 const path = await getProfileDataPath($newProfile.name)
                 const machineId = await Platform.getMachineId()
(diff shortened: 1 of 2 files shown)


packages/shared/routes/setup/Profile.svelte
```diff
@@ -68,7 +68,7 @@
 
                 const path = await getProfileDataPath($newProfile.name)
                 const machineId = await Platform.getMachineId()
-                const { sendCrashReports } = $initAppSettings
+                const { sendCrashReports } = $initAppSettings ?? {}
```
maxwellmattryan, line 71: I think it may be better to default to `false` here rather than an empty object `{}` since it is just a boolean option. I believe there's more safety defaults down the line but it could also be done here for clarity.

amadeu2: Thanks! currently the error in the console is about destructuring failing, I've just tried with `false` and still getting this error, so for getting work destructuring we might need an object.

maxwellmattryan: Ah right, perhaps then `$initAppSettings ?? { sendCrashReports: false }` might work or `const sendCrashReports = $initAppProfile?.sendCrashReports || false`? Probably doesn't matter as long as it's a boolean value rather than `undefined`.

amadeu2: yeah both options looks better, will do. Yes it's only a temporal fix meanwhile we add Sentry to mobile, would be nice to have it for alpha release to better identify errors

maxwellmattryan (approved):
LGTM!

Issue #2476, opened 2022-03-04
[Task]: Refactor router into a more generic component


Tuditi:
### Task description

There is a clear need for a generic router which contains the state machines for the navigation of the views. The goal is to have a clear way of handling navigation in our application to enable a global back button (cfr. Slack/Spotify).

This affects the following state machines
- AppRouter
- AccountRouter,
- LedgerRouter,
- SettingsRouter,
- WalletRouter,

### Requirements

Navigation on all these components work as before
A global back is implemented

### Acceptance criteria

TBD

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #2505, opened 2022-03-07
chore: refactor router

Tuditi:
## Summary

Refactoring the router

### Changelog
```
Please write an accurate and concise changelog for your changes.
```

## Relevant Issues

closes: https://github.com/iotaledger/firefly/issues/2476

## Type of Change

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions

## Checklist


packages/desktop/electron/lib/menu.js
```diff
@@ -1,5 +1,5 @@
 import { app, ipcMain, Menu, shell } from 'electron'
-import { WalletRoutes, ExternalRoute } from 'shared/lib/typings/routes'
+import { WalletRoutes, ExternalRoute } from '@core/router/enum/routes'
```
maxwellmattryan, line 2: Just FYI, the modules should contain the barrel files so that this line can be either what it is now _or_ just `import { ... } from '@core/router'`. I think we should try and import code from the highest-level directory possible and only specify sub folders of a module when we absolutely have to, like within a module itself that needs to reference its own types or just to avoid circular imports.

maxwellmattryan: I suppose the case here is that we _can_ specify a specific location as that's all that's being used here.

packages/shared/lib/core/index.ts
```diff
@@ -1 +1 @@
-export default {}
+export * from './router/'
```
maxwellmattryan, line 1: It might be better to name the export here once other exports are added, e.g. `export * as router from './router`. It should work anyways with the proper barrel file setup, but I have a weird preference to explicitly name it :sweat_smile:.

maxwellmattryan (approved):
Lots of code but LGTM :slightly_smiling_face:

packages/desktop/App.svelte
```diff
@@ -44,7 +36,6 @@
         Splash,
         Welcome,
     } from 'shared/routes'
-    import { get } from 'svelte/store'
```
nicole-obrien, line 47: We needed this 😢

Issue #2162, opened 2022-01-17
[Task]: (Wallet Tab) Add warnings to shimmer and assembly asset tiles


nicole-obrien:
### Task description

Move the relevant warnings from develop (from account tile) to single-account branch (to asset tile). Only add corresponding warnings (tooltip) to specific asset.

### Requirements

TBD

### Acceptance criteria

TBD

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #2603, opened 2022-03-23
feat: rewards per account and staking warnings

### LINKED ISSUE #2161: [Task]: (Wallet Tab) Make rewards displayed in the asset tiles per account instead all accounts
### Task description

TBD

### Requirements

TBD

### Acceptance criteria

TBD

### Creation checklist

jeeanribeiro:
## Summary
Adds rewards per account and staking warnings to asset tiles because the code needed to be adapted to single wallet refactor

## Relevant Issues
closes #2161 #2162 

## Type of Change
Please select any type below that applies to your changes, and delete those that are not.

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions
Needs to be tested with all staking possible states

## Checklist

### DIFF
--- a/packages/shared/components/AssetTile.svelte
+++ b/packages/shared/components/AssetTile.svelte
@@ -1,35 +1,46 @@
 <script lang="typescript">
-    import { Icon, Text } from 'shared/components'
+    import { Icon, Text, StakingAssetTile } from 'shared/components'
     import { isBright } from 'shared/lib/helpers'
-    import { Asset } from 'shared/lib/typings/assets'
+    import { Asset, Token } from 'shared/lib/typings/assets'
 
     export let asset: Asset
+
+    const isStakingAsset = asset?.name === Token.Assembly || asset?.name === Token.Shimmer
 
     $: assetIconColor = isBright(asset?.color) ? 'gray-800' : 'white'
 </script>
 
-<div
-    style="--asset-color: {asset?.color}"
-    class="flex flex-row justify-between items-center space-x-2 bg-gray-50 dark:bg-gray-900 p-4 rounded-2xl"
->
-    <div class="flex flex-row items-center space-x-4">
-        <div class="icon h-8 w-8 rounded-full flex items-center justify-center p-1">
-            <Icon classes="text-{assetIconColor}" icon={asset?.name?.toLocaleLowerCase()} height="100%" width="100%" />
+{#if isStakingAsset}
+    <StakingAssetTile {asset} />
+{:else}
+    <button
+        style="--asset-color: {asset?.color}"
+        class="flex flex-row justify-between items-center space-x-2 bg-gray-50 dark:bg-gray-900 p-4 rounded-2xl"
+    >
+        <div class="flex flex-row items-center space-x-4">
+            <div class="icon h-8 w-8 rounded-full flex items-center justify-center p-1">
+                <Icon
+                    classes="text-{assetIconColor}"
+                    icon={asset?.name?.toLocaleLowerCase()}
+                    height="100%"
+                    width="100%"
+                />
+            </div>
+            <div class="flex flex-col flex-wrap space-y-1">
+                <Text classes="font-semibold">{asset?.name}</Text>
+                {#if asset?.fiatPrice}
+                    <Text secondary smaller>{asset?.fiatPrice}</Text>
+                {/if}
+            </div>
         </div>
-        <div class="flex flex-col flex-wrap space-y-1">
-            <Text classes="font-semibold">{asset?.name}</Text>
-            {#if asset?.fiatPrice}
-                <Text secondary smaller>{asset?.fiatPrice}</Text>
+        <div class="flex flex-col flex-wrap space-y-1 text-right">
+            <Text classes="font-semibold">{asset?.balance}</Text>
+            {#if asset?.fiatBalance}
+                <Text secondary smaller>{`≈ ${asset?.fiatBalance}`}</Text>
             {/if}
         </div>
-    </div>
-    <div class="flex flex-col flex-wrap space-y-1 text-right">
-        <Text classes="font-semibold">{asset?.balance}</Text>
-        {#if asset?.fiatBalance}
-            <Text secondary smaller>{`≈ ${asset?.fiatBalance}`}</Text>
-        {/if}
-    </div>
-</div>
+    </button>
+{/if}
 
 <style type="text/scss">
     .icon {
--- a/packages/shared/components/AssetTile.svelte
+++ b/packages/shared/components/AssetTile.svelte
@@ -13,7 +13,7 @@
 {#if isStakingAsset}
     <StakingAssetTile {asset} />
 {:else}
-    <button
+    <div
         style="--asset-color: {asset?.color}"
         class="flex flex-row justify-between items-center space-x-2 bg-gray-50 dark:bg-gray-900 p-4 rounded-2xl"
     >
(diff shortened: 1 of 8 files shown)


packages/shared/components/StakingAssetTile.svelte
```diff
@@ -0,0 +1,246 @@
...
+    function toggleTooltip(): void {
+        canShowTooltip = !canShowTooltip
+    }
+    function handleTileClick(): void {
+        openPopup({ type: 'airdropNetworkInfo', props: { airdrop } })
+    }
+    function _getAccount(accounts: WalletAccount[]): WalletAccount {
+        return accounts.find((account) => account.alias === $selectedAccount.alias)
+    }
+    function getLocalizedTooltipText(): TooltipText {
```
Tuditi, line 77: This function needs refactoring. It's very difficult to follow. Try writing it with only one layer of if's and without elses or closures. I think they add unnecessary complexity.

Tuditi: If you want we can have a look at it together

packages/shared/lib/assets.ts
```diff
@@ -35,17 +28,25 @@ export const assets = derived(
                 color: '#6E82A4',
             },
         ]
-        if ($assemblyStakingRewards) {
```
Tuditi, line 38: If we don't use assemblyStakingRewards (or shimmer), then please remove it. Another approach is to create a store that only stores the rewards of the slected account under `stores/participation.ts`, since we won't need a total anymore with the single account refactor. Now we're just leaving dead code in the files.

Tuditi (changes requested):

packages/shared/components/StakingAssetTile.svelte
```diff
@@ -0,0 +1,246 @@
...
+                        localize('tooltips.stakingMinRewards.bodyDidNotReachMin', {
+                            values: {
+                                airdrop: capitalize(airdrop),
+                                airdropRewardMin: minimumRewards,
+                            },
+                        }),
+                    ],
+                }
+            } else if (!isAccountStaked($selectedAccount?.id) && isStakingPossible($stakingEventState)) {
+                const timeNeeded = <number>getTimeUntilMinimumAirdropReward($selectedAccount, airdrop)
```
Tuditi, line 100: This smells like getTimeUntilMinimumAirdropReward couldn't make a decision to return a number only. Why is this? Maybe it needs changing. I think functions should know whether to return a string or a number, it's their responsibility, not yours to convert the type

packages/shared/components/StakingAssetTile.svelte
jeeanribeiro: I've copied [this code](https://github.com/iotaledger/firefly/blob/develop/packages/shared/components/AccountTile.svelte#L110-L222) and done some refactoring, but yeah I will improve it more

packages/shared/components/StakingAssetTile.svelte
```diff
@@ -0,0 +1,246 @@
...
+        (isBelowMinimumRewards && !_getAccount($stakedAccounts) && isStakingPossible($stakingEventState)) ||
+        (isBelowMinimumRewards && hasStakingEnded)
+
+    function toggleTooltip(): void {
+        canShowTooltip = !canShowTooltip
+    }
+    function handleTileClick(): void {
+        openPopup({ type: 'airdropNetworkInfo', props: { airdrop } })
+    }
+    function _getAccount(accounts: WalletAccount[]): WalletAccount {
```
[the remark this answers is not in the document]
jeeanribeiro: But there is a difference between them, one is used for all accounts list in profile (Transaction) and the other is used for stakedAccounts and partiallyStakedAccounts (StakingAssetTile), also they use different props for each comparison

packages/shared/components/AssetTile.svelte
```diff
@@ -1,35 +1,46 @@
...
-    style="--asset-color: {asset?.color}"
-    class="flex flex-row justify-between items-center space-x-2 bg-gray-50 dark:bg-gray-900 p-4 rounded-2xl"
->
-    <div class="flex flex-row items-center space-x-4">
-        <div class="icon h-8 w-8 rounded-full flex items-center justify-center p-1">
-            <Icon classes="text-{assetIconColor}" icon={asset?.name?.toLocaleLowerCase()} height="100%" width="100%" />
+{#if isStakingAsset}
+    <StakingAssetTile {asset} />
+{:else}
+    <button
```
begonaalvarezd, line 16: Why did we convert this to a button?

packages/shared/components/StakingAssetTile.svelte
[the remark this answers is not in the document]
Tuditi: You're right. _getAccount() is not useful anymore, since we could just use selectedAccount in Transaction.svelte

packages/shared/components/AssetTile.svelte
begonaalvarezd: I would keep the HTML we had before and wouldnt make the assets clickable

packages/shared/lib/participation/stores.ts
```diff
@@ -207,6 +207,24 @@ export const shimmerStakingRewards: Readable<number> = derived(participationOver
     sumStakingRewards(StakingAirdrop.Shimmer, overview)
 )
 
+/**
+ * The total accumulated rewards for selected account
+ */
+export const selectedAccountStakingRewards: Readable<{ assembly: number; shimmer: number }> = derived(
```
begonaalvarezd, line 213: This has to depend on the selected account also, otherwise it will not work properly

packages/shared/lib/participation/stores.ts
```diff
@@ -207,6 +207,24 @@ export const shimmerStakingRewards: Readable<number> = derived(participationOver
...
+    participationOverview,
+    (overview) => {
+        const selectedAccountOverview = overview.find(
+            (accountOverview) => accountOverview.accountIndex === get(selectedAccountId)
```
begonaalvarezd, line 217: Will this ever find? accountIndex is an index number, and selectedAccountId is the string ID of the account, so I dont think this can work

begonaalvarezd: imo this store should be
```ts
export const selectedAccountStakingRewards: Readable<{ assembly: number; shimmer: number }> = derived(
    [participationOverview, selectedAccountId],
    ([$participationOverview, $selectedAccountId]) => {
        const selectedAccountOverview = $participationOverview.find(
            (accountOverview) => accountOverview.accountIndex === get(selectedAccount)?.index
        )
        if (selectedAccountOverview) {
            return {
                assembly: selectedAccountOverview.assemblyRewards + selectedAccountOverview.assemblyRewardsBelowMinimum,
                shimmer: selectedAccountOverview.shimmerRewards + selectedAccountOverview.shimmerRewardsBelowMinimum,
            }
        }
    }
)
```
but it shouts a circular dependency, so we need to move this around 🥲

begonaalvarezd: Im not making it derived depending on `selectedAccountId` so it only triggers when the ID changes, and not the account itself which can change a lot while selected, but we can also make this derived store dependent of `selectedAccount`

packages/shared/components/StakingAssetTile.svelte
jeeanribeiro: Will refactor it later, we need this PR merged ASAP

begonaalvarezd (approved):

PR #2902, opened 2022-04-14
fix: attempt rename profile folder everytime app is run

nicole-obrien:
## Summary
Please summarize your changes, describing __what__ they are and __why__ they were made.
This is needed because in 1.4.+ we rename the profile folders to use the profile index, this means that we can freely name and rename profiles. However, this store is not cleared when downgrading versions so this can only occur once.

in 1.4.0-1.4.2 there are bugs that meant people where downgrading to 1.3.3, because the folders had been renamed they where cleared and in 1.3.3 the user had to make there profiles again. If they then upgrade to 1.5.0, the profiles will wipe again as they are using 1.3.3- naming convention. This change simply attempts to rename profiles everytime the app loads in 1.5.0 so that people are able to upgrade.

We can fix this better or change it back once the majority of users are on a stable build of 1.5.0+

### Changelog
```
- change condition to rename folders on mac to always true on launch.
```

## Relevant Issues

## Type of Change

### Platforms
- __Desktop__
	- [x] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions

## Checklist

### DIFF
--- a/packages/shared/lib/profile.ts
+++ b/packages/shared/lib/profile.ts
@@ -468,7 +468,7 @@ async function renameProfileFolder(oldName: string, newName: string): Promise<vo
 export async function renameOldProfileFoldersToId(): Promise<void> {
     const walletPath = await getWalletDataPath()
     const profileFolders = await Platform.listProfileFolders(walletPath)
-    const oldProfiles = get(profiles).filter((profiles) => profileFolders.find((p) => p === profiles.name))
+    const oldProfiles = get(profiles).filter((profile) => profileFolders.find((p) => p === profile.name))
 
     if (oldProfiles.length > 0) {
         await Promise.all(
--- a/packages/shared/lib/profile.ts
+++ b/packages/shared/lib/profile.ts
@@ -465,11 +465,16 @@ async function renameProfileFolder(oldName: string, newName: string): Promise<vo
     await Platform.renameProfileFolder(oldPath, newPath)
 }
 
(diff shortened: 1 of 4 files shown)


Tuditi (approved):
I changed the code, so that upon startup all the old profiles are renamed to ID folders.

nicole-obrien:
> I changed the code, so that upon startup all the old profiles are renamed to ID folders.

Thank you thats much better 😄  I will retest soon

packages/shared/lib/profile.ts
```diff
@@ -465,11 +465,16 @@ async function renameProfileFolder(oldName: string, newName: string): Promise<vo
...
-    const _profiles = get(profiles)
-    await Promise.all(
-        _profiles.map(async (profile) => {
-            await renameProfileFolder(profile.name, profile.id)
-        })
-    )
+export async function renameOldProfileFoldersToId(): Promise<void> {
+    const walletPath = await getWalletDataPath()
+    const profileFolders = await Platform.listProfileFolders(walletPath)
+    const oldProfiles = get(profiles).filter((profiles) => profileFolders.find((p) => p === profiles.name))
```
rajivshah3, line 471:
```suggestion
    const oldProfiles = get(profiles).filter((profile) => profileFolders.find((p) => p === profile.name))
```

PR #3049, opened 2022-04-26
feat: basic transaction details popup

nicole-obrien:
## Summary
Please summarize your changes, describing __what__ they are and __why__ they were made.

### Changelog
```
Please write an accurate and concise changelog for your changes.
```

## Relevant Issues

## Type of Change

### Platforms
- __Desktop__
	- [ ] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions

## Checklist


packages/shared/components/atoms/KeyValueBlock.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="typescript">
+    import { FontWeightText } from 'shared/components/Text.svelte'
```
maxwellmattryan, line 2: It would be nice to start creating a library specifically for the UI, then we could just import `FontWeightText` from there. There's some functions in `shared/lib/utils` at the moment that could also be moved, plus I'm sure there's some other logic elsewhere that really only applies to UI. Will give this a think :thinking:

packages/shared/components/atoms/pills/ActivityTypePill.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="typescript">
+    import { localize } from '@core/i18n'
+    import { isParticipationPayload } from '@lib/wallet'
+    import Pill from './Pill.svelte'
+
+    export let activity
+    export let confirmed
+
+    enum ActivityType {
```
maxwellmattryan, line 9: I think `ActivityType` isn't quite descriptive enough, maybe `TransactionActivityType` or something?

packages/shared/components/atoms/pills/ActivityTypePill.svelte
```diff
@@ -0,0 +1,68 @@
...
+        Received = 'received',
+        Sent = 'sent',
+        Receiving = 'receiving',
+        Sending = 'sending',
+    }
+
+    let type: ActivityType
+    $: {
+        if (activity) {
+            if (isParticipationPayload(activity)) {
```
maxwellmattryan, line 22: It seems weird to be passing an `activity` to the `isParticipationPayload` function imo. Perhaps there is a better / more descriptive word than `activity`?

packages/shared/components/popups/ActivityDetailsPopup.svelte
```diff
@@ -0,0 +1,91 @@
...
+        } else if (milestonePayload) {
+            value = getMilestoneMessageValue(milestonePayload, $accounts)
+        } else if (transactionPayload) {
+            value = transactionPayload.data.essence.data.value
+        }
+    }
+    $: currencyValue = convertToFiat(
+        value,
+        $currencies[CurrencyTypes.USD],
+        $exchangeRates[$activeProfile?.settings.currency]
```
maxwellmattryan, line 54: Might be good to add a fallback to USD here, i.e. `$exchangeRates[$activeProfile?.settings.currency ?? CurrencyTypes.USD]`.

packages/shared/components/atoms/pills/ActivityTypePill.svelte
nicole-obrien: But a transaction is a type of activity, in the future when we have more types of activities that transactions :)

packages/shared/components/popups/ActivityDetailsPopup.svelte
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    $: transactionSubjectAddressOrAccount = getTransactionSubjectAddressOrAccount(transactionPayload)
+</script>
+
+<transaction-details class="w-full h-full space-y-6 flex flex-auto flex-col flex-shrink-0">
```
maxwellmattryan, line 60: I haven't seen this element usage before, what is `<transaction-details ...>` Should it be `TransactionDetails` or are you doing something different here?

packages/shared/components/atoms/pills/ActivityTypePill.svelte
nicole-obrien: Yeah its because I was renaming things to activity within this portion of code, because this is part of the ActivityDetailPopop openable by a ActivityCard. Also for now just trying to update UI and maintain backwards compatibility with existing IOTA before we refactor some of the logic for stardust.

packages/shared/lib/utils/transactionObject.ts
```diff
@@ -0,0 +1,33 @@
+import { Payload } from '@lib/typings/message'
```
maxwellmattryan, line 1: If you're creating a new utils module, let's go ahead and put it in `core`, so `core/utils/transaction-object.ts` along with a barrel export file for it. Anything in these folders should use the `kebab-case` naming convention for filenames.

Arguably this belongs in a `wallet` module within `common` (esp. since you're importing a lot of functions from `wallet.ts`). I've mentioned to David how it'd be helpful having `wallet.ts` split up specific modules, e.g.
- `account` - retrieves account data, syncs accounts, handles addresses, handles transfers, account name, etc.
- `network` - handles anything network related like setting the client options, getting node info
- `stronghold` - handles backups, any software security-related things like mnemonic creation and verification, locking Stronghold, etc.
- `ledger` - handles specific Ledger-related stuff (not much really besides querying for the device status)
- `participation` - handles everything for voting and staking
- `migration` - handles everything for migration
- `storage` - handles storage stuff, i.e. managing PINs, encrypting / decrypting the storage, deleting storage

... and I think this util function would belong somewhere in there. We agreed to start doing this after the new bindings are up to speed.

packages/shared/components/popups/ActivityDetailsPopup.svelte
nicole-obrien: It's just a custom element with no other use than acting as a container that is named for easy organisation and reading. At least I think and not aware of any specific side affects if not using specific element attributes with it.

packages/shared/lib/utils/transactionObject.ts
nicole-obrien: I can move it and follow the conventions, my only concern is its not really a library function per say for managing an object or context of the wallet. just a helper function so not sure it belongs in core?

packages/shared/components/atoms/KeyValueBlock.svelte
nicole-obrien: Yeah, this was part of another PR for the text component, but because its the props for this component I kept it there. Problem is even typing exported props in svelte doesn't give you nice to use prefills in the IDE, so I thought I would expose the enum so we don't need to worry about typing it in.

packages/shared/routes/dashboard/wallet/views/AccountHistory.svelte
```diff
@@ -25,7 +25,10 @@
     export let transactions: AccountMessage[] = []
```
maxwellmattryan, line 25: This isn't a part of your changes, but we currently mix the concept of "transactions" and "messages" :sweat:. We should keep this in mind while we're transitioning to the new wallet-rs changes that revolving around UTXOs and not protocol messages anymore (hopefully should be simpler for us).

nicole-obrien: Yeah, I am on my third branch of nested changes and been refactoring stuff in this one too. For now I have changed the use of transactions with messages here. But its in a third PR ;) and will likely be refactored when we do wallet.rs changes.

packages/shared/components/atoms/pills/ActivityTypePill.svelte
maxwellmattryan: That makes sense, although maybe we can rename the function in that case?

nicole-obrien: Stardust will not have the participations for now. Would be good to refactor when its added. I imagine in a few PRs time I will remove this function call for now 👍

packages/shared/components/popups/ActivityDetailsPopup.svelte
nicole-obrien: Perhaps, this currently only works for IOTA and we need to rethink this anyway. Also this code already exists on develop.

packages/shared/lib/utils/transactionObject.ts
maxwellmattryan: Yeah that's true, would be another good one for a library of specific UI functions.

maxwellmattryan (approved):
LGTM!

PR #3234, opened 2022-05-17
fix, refactor: Stronghold import screen

rajivshah3:
## Summary
On desktop and mobile:
- The `<input>` element covers the "Continue" button, so if the user tries to click it, the file picker is repeatedly shown when importing a Stronghold
- Removed some conditional rendering
- Fixed typo: `extentions` -> `extensions`

On mobile only:
- Matched flow of desktop (import, then user must click "Continue")
- Removed `<input>` element as its behavior was slightly different from calling the native file picker
- Get file name by splitting on the path separator instead of using a regex (the current regex would not work with file names that included spaces, such as `Firefly Test Account.stronghold`)

### Changelog
```
- Removed duplicate <input> element so "Continue" button is not covered
- Removed conditional rendering
- Matched mobile flow to desktop flow
- Fixed getting file name from the file path
- Fixed typo
```

## Relevant Issues
N/A

## Type of Change

### Platforms
- __Desktop__
	- [x] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [x] iOS
	- [ ] Android

### Instructions
Import a Stronghold when creating a new profile

## Checklist


rajivshah3:
To do: see if `<input>` element should be removed entirely, because its behavior on mobile differs from calling `handleFileSelectMobile`

amadeu2 (commented):
On desktop all works well as you described, but on mobile I've tested on Android and can't be changed the selection once done, we have to come back and forth again to make a new selection. In the next screen (the password one) the back button stop to work.

I think we should change the `Drag and drop` wording by something specifiacally for mobile, but also there is a new design flow done for mobile by the UX team to be implemented probably next week.

Maybe wait for the user to tap the `continue` button could feels like an extra step imo. Right now on `develop` branch is not working yet cause there is needed some merges more, but on `test/mobile-alpha` the user can go back on the password screen and choose another file again.

rajivshah3:
> On desktop all works well as you described, but on mobile I've tested on Android and can't be changed the selection once done

Ooh you're right. This is actually the current behavior on desktop too, I guess I never realized 🤔 

> In the next screen (the password one) the back button stop to work.

This is an issue on desktop too. I'll investigate and see if I can fix it

> Maybe wait for the user to tap the continue button could feels like an extra step imo. Right now on develop branch is not working yet cause there is needed some merges more, but on test/mobile-alpha the user can go back on the password screen and choose another file again.
But yes, the users could like it, we could test with continue button and also propose to the UX team for the next implementation.

Yep, maybe we should run this by the UX team then. My primary aim was to fix the `<input>` element on this screen because it effectively prevents the user from proceeding with the Stronghold import (on both desktop and mobile), but I saw some areas where we could align the flow with desktop and refactor the component to be simpler. I'm fine with reverting those changes as long as we can fix the `<input>` issue 👍

rajivshah3:
> In the next screen (the password one) the back button stop to work.

Addressed this in #3288 since the issue is coming from a different component

amadeu2 (approved):
Great, lets merge this change for show to the UX team 💪

PR #3381, opened 2022-06-01
chore: make Capacitor setup easier and CI-friendly

rajivshah3:
## Summary
This PR combines `capacitor-sample.config.ts` and the gitignored `capacitor.config.ts`. `capacitor.config.ts` used to contain developer-specific settings such as the local IP, so it was not checked in to Git. By putting it in Git and using environment variables, we can ensure that changes to things such as plugin configurations are mirrored to all developers, but developers can still set their local IP without having to change a file that is under version control. This also makes the workflow easier for CI, because `capacitor.config.ts` would need to be modified with `jq` or `sed` otherwise.

### Changelog
```
- Use environment variables to set Capacitor server
- Update docs
```

## Relevant Issues
N/A

## Type of Change
- [x] __Docs__ - changes to the documentation

### Platforms
- __Desktop__
	- [x] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [x] iOS
	- [ ] Android

### Instructions
Verified that `capacitor.config.json` is generated properly when using environment variables with `yarn cap copy`

## Checklist

### DIFF
--- a/packages/mobile/.gitignore
+++ b/packages/mobile/.gitignore
@@ -1,4 +1,3 @@
-capacitor.config.ts
 capacitor.config.json
 capacitor.plugins.json
 capacitor.settings.gradle
--- a/packages/mobile/README.md
+++ b/packages/mobile/README.md
@@ -48,14 +48,18 @@ packages/mobile $ yarn ios:update
 ### Development
 
 Capacitor Live Reload setup:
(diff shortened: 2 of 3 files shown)


amadeu2 (approved):
great, nice to have! 😀

PR #3385, opened 2022-06-03
feat: update asset dropdown list

### LINKED ISSUE #3326: [Subtask]: Update asset dropdown to allow the user to select from list of assets

jeeanribeiro:
## Summary
Updates asset dropdown list to have changeable assets in send form popup

## Relevant Issues
closes #3326

## Type of Change
- [x] __New__ - a change that implements a new feature

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

## Checklist


maxwellmattryan (approved):
LGTM!

nicole-obrien (changes requested):
Not a fan on the styling, lets take this to UX to flesh out.

packages/shared/lib/core/wallet/utils/formatTokenAmountDefault.ts
```diff
@@ -5,6 +5,6 @@ export function formatTokenAmountDefault(amount: number, tokenMetadata: ITokenMe
     return (
         formatNumber(amount / 10 ** tokenMetadata?.decimals, 0, tokenMetadata?.decimals, 0, true) +
         ' ' +
-        tokenMetadata.unit
+        tokenMetadata?.unit
```
MarkNerdi, line 8: If there is a case where this is `undefined` then it will append that to the string (e.g. 100 undefined). Is that desired?

jeeanribeiro: Not desired, I've fixed it :smile:

packages/shared/components/inputs/AssetDropdown.svelte
```diff
@@ -1,32 +1,80 @@
...
-    import { FontWeightText } from 'shared/components/Text.svelte'
-    import { Icon, Text } from 'shared/components'
     import { assets } from '@core/wallet'
+    import { AssetTile, Icon, Text } from 'shared/components'
+    import { FontWeightText } from 'shared/components/Text.svelte'
+    import { clickOutside } from 'shared/lib/actions'
 
     export let asset = $assets?.[0]
+
+    const hasMultipleAssets = $assets.length > 1
```
MarkNerdi, line 9: should this be following:
`$: hasMultipleAssets = $assets.length > 1` ?

MarkNerdi (commented):
One small comment. The rest looks good for me

packages/shared/components/inputs/AssetDropdown.svelte
begonaalvarezd: good catch, thanks!

nicole-obrien (changes requested):
Few tiny polishes:
- The tile text on the left is aligned on the left, and vice versa the tile text on the right is aligned on the right.
- Also I think the vertical spacing between these list items is too big.
- Change the max height so we can fit 3/4/5 items in it depending on screen size? 3 at a minimum before there is a scroll bar?

PR #3386, opened 2022-06-03
refactor: networking logic

### LINKED ISSUE #3319: [Task]: Update network config settings for stardust
### Task description

We need to update all the network config settings to ensure that they work with the new stardust api and network module. 
For now I think we can ignore the concept of a primary node. And the user can no longer switch an account so we can clean that up.

### Tech Specs

1. [x] Allow one client per profile (i.e. only one set of `IClientOptions`)
2. [x] Allow client configuration in the network config settings 
4. [x] Allow view node info in network config settings
5. [x] Refactor `NodeInfoPopup` component, change to individual copyable boxes
6. [x] Disallow network switching for a profile
8. [x] Replace concept / code of `networkConfig` to be `clientOptions`
7. [ ] (Opt.) Refactor existing network logic (potentially create `client` module)
3. [ ] ~~(Opt.) Allow client configuration during onboarding (node settings)~~

### Requirements

- [x] Hide "Node configuration" when automatic node selection is selected
- [x] Fix "Node configuration" list for all cases
- [x] Fix "Add node" click

### Acceptance criteria

Meets the above requirements.

### Creation checklist

maxwellmattryan:
## Summary
With the new changes to the bindings, the app-side code needed to be refactored / updated.

### Changelog
```
TBD
```

## Relevant Issues
Closes #3319

## Type of Change
- [x] __Update__ - a change which updates existing functionality

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions
Tests network configuration settings logic.

## Checklist


packages/shared/lib/utils.ts
```diff
@@ -476,3 +476,12 @@ export function range(size: number, start: number = 0): number[] {
...
     return formatDate(date, { year: 'numeric', month: 'short' })
 }
+
+/*
+ * Gets an object value from an object string path
+ */
+export function resolveObjectPath(object: unknown, path: string, defaultValue: unknown): unknown {
+    if (!path) return
+
+    return path.split('.').reduce((o, p) => (o ? o[p] : defaultValue), object)
```
rajivshah3, line 486: I think users of this function might expect behavior similar to `lodash.get`, which resolves the whole path and then returns `defaultValue` if the value at the path is `undefined`. This seems like it will replace an intermediate object during a `reduce` operation with the `defaultValue`, which may be incorrect behavior in some cases. For example, in this [JSFiddle](https://jsfiddle.net/mjbqL29d/1/), `undefined` is returned instead of the `defaultValue`. Based on this [plain JS version](https://youmightnotneed.com/lodash#get), I think this should work:

```suggestion
    const result = path.split('.').reduce((o, p) => (o && o[p]), object)
    return result === undefined ? defaultValue : result
```

rajivshah3 (changes requested):

packages/shared/components/popups/ReceivePopup.svelte
```diff
@@ -16,7 +16,7 @@
     <div class="mb-6 flex flex-col items-center">
         <Text secondary classes="mb-6 text-xs">
             {localize('general.myNetworkAddress', {
-                values: { networkName: $activeProfile?.settings.networkConfig.network.name },
```
nicole-obrien, line 19: Just realised this file isn't used and we can delete it

packages/shared/components/modals/NetworkSummaryModal.svelte
```diff
@@ -15,7 +15,7 @@
 <Modal bind:this={modal} position={{ left: '80px', top: '25px' }}>
     <network-indicator-content class="flex flex-col">
         {#if $activeProfile?.isDeveloperProfile}
-            <Text type="h3" classes="px-7 pt-5">{$activeProfile?.settings.networkConfig.network.name}</Text>
+            <Text type="h3" classes="px-7 pt-5">{$activeProfile?.settings.clientOptions.network}</Text>
```
nicole-obrien, line 18: This is never set locally so it is undefined, we probably should get the network from the node info.

packages/shared/components/Profile.svelte
```diff
@@ -34,9 +34,8 @@
 
 <div class="flex items-center justify-center w-24">
     <div class="flex flex-col justify-between items-center">
-        <div class="relative mb-3">
+        <div on:click={handleOnClick} class="relative cursor-pointer mb-3">
             <div
-                on:click={handleOnClick}
                 class="h-18 w-18 {bgColor
```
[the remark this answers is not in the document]
jeeanribeiro: It's trying to solve the problem of clicking on the network icon that is over the profile icon doing nothing, and it should be clickable to login as well

packages/shared/components/popups/NodeInfoPopup.svelte
```diff
@@ -0,0 +1,172 @@
...
+        {/key}
+    </div>
+    {#each Object.keys(NODE_INFO_TAB_MAP[nodeInfoTab]).map( (nodeInfoTabKey) => processNodeInfoMapTab(nodeInfoTab, nodeInfoTabKey) ) as { localeKey, nodeInfoValue }}
+        <CopyableBox value={nodeInfoValue} classes="mb-2.5">
+            <div class="w-full h-full flex flex-row justify-between items-center">
+                <Text type="p" fontSize="sm" fontWeight="font-600" secondary
+                    >{localize(`${NODE_INFO_LOCALE_BASE_PATH}.${localeKey}`)}</Text
+                >
+                {#if typeof nodeInfoValue === 'boolean'}
+                    <Checkbox disabled checked={nodeInfoValue} />
```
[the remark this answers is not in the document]
jeeanribeiro: baseToken.useMetricPrefix, it is not showing only when the value is false because of a Copyable box, condition, but will be fixed

laumair:
Clicking on `Add Node` breaks the app. 

Screenshot 2022-06-13 at 16 48 53

laumair (approved):

PR #3397, opened 2022-06-07
feat: add image qr generator

amadeu2:
## Summary
Current behavior:
The QR Code generator creates an SVG wich contains too many nodes to be used inside the Drawer component on mobile. Testers reported that the Drawer opening animation takes up to 5 seconds and hence it tends to an erratic behavoir.

This PR uses a new QR Code generator which uses an image as a css background for just one HTML element, resulting in a smooth Drawer animation.

### Changelog
```
- It was made a clone of the `qrcode-generator` repository and implemented background and foreground options to use in dark mode.
- Implemented a new component to generate a QR Code as image css background.
- Use that component on the Receive drawer.
```

## Type of Change

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [x] iOS
	- [x] Android

## Checklist


amadeu2:
As we have problems scanning inverted QR Codes with the current Capacitor QR scan plugin, we'll keep only the light mode version by now. So I hidded the related code.

packages/shared/components/QRImage.svelte
```diff
@@ -0,0 +1,43 @@
...
+<script lang="typescript">
+    import qrcode from 'qrcode-generator'
+    import { appSettings } from 'shared/lib/appSettings'
+
+    export let data = ''
+    export let size = 6
+
+    let qr: QRCode
+    let QRTag = ''
+    let background = ''
```
paul-boegelsack, line 17: const instead of let?

packages/shared/components/QRImage.svelte
```diff
@@ -0,0 +1,43 @@
...
+    let foreground = ''
```
paul-boegelsack, line 18: const instead of let?

paul-boegelsack (approved):
Looks good to me. Just two minor things that should be addressed because of warnings.

packages/shared/components/QRImage.svelte
amadeu2: fixed thanks!

PR #3399, opened 2022-06-07
fix: init performance on mobile

amadeu2:
## Summary
Current behavior:
1- The splash screen hides before locales are loaded, notably visible at loading them.
2- There is a type mismatch when loading the css main file. Also based on this info:
 - https://web.dev/preload-scanner/
 - https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload
seems that there is room for improvment.

This PR uses the `tick` Svelte function to wait until the locales are applied to finally hide the splash screen.
Also removes the `type` property on the `link` tag.
Lastly I added the `link` tags with `preload` value for CSS and JS main files ["...to start loading early in the page lifecycle, before browsers' main rendering machinery kicks in. This ensures they are available earlier and are less likely to block the page's render, improving performance."](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload)

### Changelog
```
- Fix locales sentence and add tick function.
- Add `link` tags for CSS and JS main resources.
```

## Type of Change

### Platforms
- __Desktop__
	- [ ] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [x] iOS
	- [x] Android

## Checklist

### DIFF
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -45,7 +45,7 @@
     }
 
     $: if ($isLocaleLoaded) {
-        hideSplashScreen()
+        void hideSplashScreen()
     }
 
     async function hideSplashScreen() {
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -1,6 +1,6 @@
 <script lang="typescript">
     import { nativeSplash } from 'capacitor/capacitorApi'
-    import { onMount } from 'svelte'
+    import { onMount, tick } from 'svelte'
     import { QRScanner, Route, ToastContainer, Popup } from 'shared/components'
     import { popupState } from 'shared/lib/popup'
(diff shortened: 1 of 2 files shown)


packages/mobile/App.svelte
```diff
@@ -44,7 +44,14 @@
         document.dir = $localeDirection
     }
 
-    $: $isLocaleLoaded, nativeSplash.hide()
+    $: if ($isLocaleLoaded) {
+        hideSplashScreen()
+    }
```
paul-boegelsack, line 49: I guess, you have to add a at least a catch statement to handle the warning?

paul-boegelsack (approved):
Interesting articles! 🚀

packages/mobile/App.svelte
amadeu2: 🤔 I can't see a warning, where is it?

PR #3402, opened 2022-06-08
updated instructions for running firefly with the Speculos simulator

shufps:
## Summary
Updated instructions for letting Firefly run with the Speculos simulator


cvarley100:
LGTM. Thanks @shufps 🙌

PR #3506, opened 2022-06-14
fix: adds truncate to profile name

### LINKED ISSUE #1721: [UI] Profile names may obscure other profile names
### Issue description

If the profile name is too long it may obscure other profile names in the profile overview/login window.

Bildschirmfoto 2021-12-13 um 16 01 54

### Firefly version

1.3.0

### OS details

macOS 12.0.1

### Can the issue be reliably reproduced?

Yes

1. Create a new profile with close to 20 characters 
2. Log out
3. Enjoy

### Expected behaviour

The profile overview/login window should possibly only show the first x characters of the profile name, followed by three dots "...".

### Actual behaviour

The profile overview/login window shows the whole profile name and obscures ui to its left and right.

### Errors

_No response_

jeeanribeiro:
## Summary
Adds truncate and fixed width for profile name because if the name is too big it will obscure others in the Profile selection screen

## Relevant Issues
closes #1721

## Type of Change
- [x] __Fix__ - a change that fixes an issue

### Platforms
- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

## Checklist


packages/shared/components/Profile.svelte
```diff
@@ -53,7 +53,7 @@
                     height={14}
                 />
             {/if}
-            <Text type="h5" classes="text-center">{name}</Text>
+            <Text type="h5" classes="text-center w-28 truncate">{name}</Text>
```
begonaalvarezd, line 56: Tiny comment here:

I think it would be more flexible if instead of hardcoding here a new width, you inherit from the top father which already has `w-24`. If you pass `w-full` to the 2 node parents of this Text component, it will automatically adapt to the available space from the father, so if we set the top width to `w-40`, it will automatically work

image

packages/shared/components/Profile.svelte
```diff
@@ -32,8 +32,8 @@
     }
 </script>
 
-<div class="flex items-center justify-center w-24">
-    <div class="flex flex-col justify-between items-center">
+<div class="flex items-center justify-center w-40">
```
begonaalvarezd, line 35:
```suggestion
<div class="flex items-center justify-center w-24">
```

begonaalvarezd (changes requested):

packages/shared/components/Profile.svelte
```diff
@@ -48,7 +48,7 @@
             </div>
             <NetworkIconBadge {networkType} {networkProtocol} />
         </div>
-        <div class="flex flex-row items-baseline space-x-1.5 mb-2">
+        <div class="flex flex-row items-baseline space-x-1.5 mb-2 w-full">
```
MarkNerdi, line 51: with this, short names are not centered anymore. Try `content-center`

MarkNerdi (changes requested):
Added 1 small comment. Also fix it please on the branch for dev

packages/shared/components/Profile.svelte
begonaalvarezd: `justify-center` seems to do the trick

begonaalvarezd: pushed the fix

begonaalvarezd (approved):
LGTM 👌🏼

PR #3585, opened 2022-06-19
fix: password key submit

### LINKED ISSUE #3582: [Subtask]: Password popup doesn't submit on press enter key

jeeanribeiro:
## Summary
Adds type button to InputContainer's button, because the default is type submit and it was always emitting a submit event when clicked.
Moves the handle submit function call to on:submit of containing form because the button of type submit tries to submit its parent form element and emits the submit event on it.

## Relevant Issues
closes #3582 

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions
- Try any action that shows the password popup
- Write the password and press the Enter key
- It should submit the password correctly

## Checklist

### DIFF
--- a/packages/shared/components/Button.svelte
+++ b/packages/shared/components/Button.svelte
@@ -22,7 +22,7 @@
     export let xsmall = false
     export let classes = ''
     export let type = 'button'
-    export let form = ''
+    export let form = undefined
     export let autofocus = false
     export let inlineStyle = ''
     export let showHoverText = undefined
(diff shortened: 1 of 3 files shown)


begonaalvarezd (changes requested):
Clikcing on the unlock button doesnt work anymore 🙏🏼

begonaalvarezd (approved):
works 🎉

PR #3684, opened 2022-06-24
feat: validate amount storage deposit before send

### LINKED ISSUE #3592: [Subtask]: Validate amount + storage deposit before send is available
We currently only validate the amount on the send form, but the storage deposit is calculated on the send confirmation screen. We should validate that the total amount that will be sent is available.

jeeanribeiro:
## Summary
Adds amount + storage deposit validation before sending and refactors the expiration time validation code

## Relevant Issues
closes #3592 

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

## Checklist


MarkNerdi (approved):

nicole-obrien:
The issue has linked properly I think we should always manually link on the right or create the branch through the issue to avoid this problem.

nicole-obrien (changes requested):
I think it is better UX if we don't allow the user to click the send button in the first place, or we add the error message inline to the send confirmation form rather than closing it and showing a notification. This will then allow the user to go back to the previous page and modify their inputs.

nicole-obrien (changes requested):
Currently getting this:
Screenshot 2022-06-28 at 11 54 30

Additionally we need to cover the generic error case or any known error cases from wallet.rs. Right now we just return nothing and show no error message if its not the two validation errors that caused the error.

jeeanribeiro:
> Currently getting this: Screenshot 2022-06-28 at 11 54 30
> 
> Additionally we need to cover the generic error case or any known error cases from wallet.rs. Right now we just return nothing and show no error message if its not the two validation errors that caused the error.

I don't know why this error is showing to you in the console, because the `InsufficientFundsForStorageDepositError` is indeed a constructor

nicole-obrien (approved):
Made a small changes to validate the send form before the password where possible, and still add error messages to the page if the send fails for another reason 👍

PR #3687, opened 2022-06-24
fix: remove duplicated close buttons on mobile popups

paul-boegelsack:
## Summary

Since the popup drawer on mobile has got a close button in general, some popups now have two close buttons.

## Changelog

```
- Removed duplicated close buttons on backup and password popup
```

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -53,7 +53,7 @@
         nativeSplash.hide()
     }
 
-    void setupI18n()
+    void setupI18n({ fallbackLocale: 'en', initialLocale: $appSettings.language })
 
     onMount(async () => {
         initRouters()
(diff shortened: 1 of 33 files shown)


amadeu2:
The close button on Password is not on Figma design. There is a cancel button but is already implemented on desktop and we sould use the same. 
We needed only a close button on Network status drawer. So I thinl Password should be restored to previous state.

amadeu2 (commented):
The close button on Password is not on Figma design. There is a cancel button but is already implemented on desktop and we sould use the same. 
We needed only a close button on Network status drawer. So I think Password should be restored to previous state. What do you think?

amadeu2 (approved):

PR #3839, opened 2022-07-08
fix: action sheet crash

### LINKED ISSUE #3782: [Task]: Fix crash at open ActionSheet
### Description

To reproduce the issue, we need to double tap the 3-dots button at account switcher drawer, before the ActionSheet plugin is opened. The app crashes with the next error:
Image

This could be better reproduced the first time the ActionSheet plugin is called. This cold-run in some cases is taking almost 1 second on my Samsung A22, that long time led me to tap more than one time. On iOS there is less time at the cold-run, but still notable on my Iphone 7 iOS 14.

### Requirements

- [ ] Works on Android
- [ ] Works on iOS

paul-boegelsack:
## Summary

This PR should fix the crash when an action sheet is called again before the last call has finished.

## Changelog

```
- Added flag to check if the action sheet has been called
```

## Relevant Issues

Closes #3782

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/shared/components/drawerContent/AccountSwitcher.svelte
+++ b/packages/shared/components/drawerContent/AccountSwitcher.svelte
@@ -33,6 +33,10 @@
 
     let toggleEdit = false
 
+    // TODO: find a better solution to avoid a crash when the action sheet is called again
+    // before the last call is finished.
+    let isActionSheetCalled = false
+
     const menuActions = [
         {
             title: localize('actions.customizeAcount'),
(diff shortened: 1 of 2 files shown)


paul-boegelsack:
Maybe there is a better way but it should fix the error.

amadeu2 (approved):
Works nice, I haven't able to crash the app again 💪
I think if we choose to endup using the Action sheet, we should clone the Capacitor repo and add this flags on Java code, also we should need to fix the dark mode. But samesly is taking some time to open the first time or cold-run, I think it could be related to the intense use of the Capacitor bindings and the Action sheet call is queued producing the delay.

PR #3885, opened 2022-07-19
fix: truncate long tx metadata

jeeanribeiro:
## Summary
Adds margin right to key text and truncate to value text in KeyValueBox

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

## Checklist


nicole-obrien (changes requested):
I think this is a good start but we should add a way for the user to view the full details? Maybe a tooltip or something?

nicole-obrien (changes requested):
I think there is two issues still:

- The little triangle of the tooltip is not showing, likely something to do with the scroll bar?
- When hovering over the tooltip, then going back over the text the flag doesn't toggle correctly and then you are left in an inverted state.

Suggestion: We should have the tooltip positioned on top and not right.

begonaalvarezd:
> I think there is two issues still:
> 
> * The little triangle of the tooltip is not showing, likely something to do with the scroll bar?
> * When hovering over the tooltip, then going back over the text the flag doesn't toggle correctly and then you are left in an inverted state.
> 
> Suggestion: We should have the tooltip positioned on top and not right.

@nicole-obrien  I pushed some changed to fix these suggestions

nicole-obrien (approved):
✅

PR #3887, opened 2022-07-19
enhancement: truncate amount transaction details

jeeanribeiro:
## Summary
Having a lot of decimals in the transaction amount breaks the layout and doesn't fit with it, so I added the truncate class when the number is too big and an ellipsis button when the amount has more than 4 decimal places, when this ellipsis is clicked it shows all decimal places

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

## Checklist


nicole-obrien (changes requested):
I think the amount of decimal places we display should be based on the total number of characters in the string. For example, we should only truncate if there is too many characters on the left and right. Lets define that number.

jeeanribeiro:
A new design was added to Figma so this PR is outdated and the best choice is to start its development from scratch again

PR #3994, opened 2022-07-25
fix: problems with account action titles for delete and hide an account

### LINKED ISSUE #3939: [Task]: Hiding an account freezes the app
### Task description

If you try to hide a wallet on android it freezes the app. Investigate the crash and fix it. 
It could be related to 1.5 and 1.6 merge changes on desktop component version.

### Requirements

- [ ] Works on Android
- [ ] Works on iOS

paul-boegelsack:
## Summary

This PR fixes the bug that causes the app to freeze when the hide action is selected. 
Additionally, it centers the title of the deleted wallet action

## Changelog

```
- Added classes attribute to hide wallet title and added necessary tailwind classes
- Centered delete wallet action title
```

## Relevant Issues

Closes #3939

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist


amadeu2 (changes requested):
I can't reproduce it 😅 where is the fix for the app crash? I can see only the css center code. 
I think I discovered another issue here. To match desktop behavior, when we create a new wallet, the other wallet should allow to be hidded and currently it shows the delete option rather than the hide option. I think it's about to the latest changes on desktop with merges and we should pair it with components at `drawerContent` folder. We could have a call to discuss it well tomorrow

paul-boegelsack:
> I can't reproduce it 😅 where is the fix for the app crash? I can see only the css center code. 

The fix is actually the css center code 😅 The tailwind classes were added to the type attribute not the classes attribute. That causes the crash, when the component has loaded. 

>  I think I discovered another issue here. To match desktop behavior, when we create a new wallet, the other wallet should allow to be hidded and currently it shows the delete option rather than the hide option.

Yeah, I discovered the same while fixing the crash: 

https://github.com/iotaledger/firefly/issues/3998

> We could have a call to discuss it well tomorrow

Sounds good, just schedule when it fits for you 👍

amadeu2:
> The fix is actually the css center code sweat_smile The tailwind classes were added to the type attribute not the classes attribute. That causes the crash, when the component has loaded.

Ahh 😄 perfect then I'll test that right now!
 
> > I think I discovered another issue here. To match desktop behavior, when we create a new wallet, the other wallet should allow to be hidded and currently it shows the delete option rather than the hide option.
> 
> I discovered the same while fixing the crash and was wondering as well:
> 
> #3998

Awesome just added it to TODO colomn, so let's merge it asap

> > We could have a call to discuss it well tomorrow
> 
> Sounds good, just schedule when it fits for you +1

maybe we could talk about it tomorrow better, thanks!

amadeu2 (approved):

PR #4008, opened 2022-07-25
feature: copy and paste for password input

### LINKED ISSUE #3803: [Task]: Fix not able to copy or paste in mobile password inputs
### Task description

In mobile, we should allow to 
- paste in any case
- copy if the password hiding is disabled (you can read the password)

### Requirements

TBD

### Acceptance criteria

TBD

### Creation checklist

paul-boegelsack:
## Summary

This PR adds the option to enable a copy/paste button for password fields. 

## Changelog

```
- Added capacitor clipboard plugin
- Added platform interface for clipboard plugin
- Added copy/paste button to password input
```

## Relevant Issues

Closes #3803

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist


paul-boegelsack:
Hey @amadeu2 I decided to create a draft request, so you can take a look. My question is, if it is okay to use the capacitor plugin? I know we have a setClipboard function in the utils but it uses deprecated logic and we do not have a paste function. I also tried navigator.clipboard but it seems not available on mobile. Maybe we can discuss the approach tomorrow in the call 🤔

amadeu2:
> Hey @amadeu2 I decided to create a draft request, so you can take a look. My question is, if it is okay to use the capacitor plugin? I know we have a setClipboard function in the utils but it uses deprecated logic and we do not have a paste function. I also tried navigator.clipboard but it seems not available on mobile. Maybe we can discuss the approach tomorrow in the call thinking

Yes let's do it, I think since it's related with password managers, maybe we could use this options:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#allowing_autocomplete
to avoid to use the capacitor plugin. I think we should not allow to be copied, the password should be generated on the password manager and then paste on the app input.

paul-boegelsack:
> Yes let's do it, I think since it's related with password managers, maybe we could use this options: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#allowing_autocomplete to avoid to use the capacitor plugin. I think we should not allow to be copied, the password should be generated on the password manager and then paste on the app input.

@amadeu2 Tried to add the autocomplete attribute (on, current-password, new-password) but without success. Maybe I do something wrong, because you said that works somewhere else? 😅 

Additionally I removed the copy function but stayed with paste, so people can at least paste there passwords form a manager manually. 

Maybe we should go with that solution for now and create a different task for the manager integration?

PR #4140, opened 2022-08-01
feat: implements pull to refresh

### LINKED ISSUE #3791: [Task]: Implement pull to refresh
Description:
https://trello.com/c/gEtroDll/32-pull-to-refresh-functionality-not-included

jeeanribeiro:
## Summary

...

## Changelog

```
```

## Relevant Issues
closes #3791 

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


amadeu2 (commented):
Also I've added a conditional on $animationValues, then opens without errors but still is not working on Android and iOS

amadeu2 (changes requested):
Sorry I forgot to update the branch 😅 I've been able to test it now. 
Android work nice 💪 the only issue is to with adding event listeners to the window, instead you could add it to a pseudo class ::after ::before or some invisible div that only respond touch events at top, not on the transaction history area to avoid conflicting with the current swipe-header-animation.
On iOS the pull down swipe action conflicts also the current swipe-header-animation, so since we have no time to fix this, we should get rid of it and use the same behavior as Android but with the iOS icon loader animation.

begonaalvarezd (commented):
Amazing work man 💪🏼 Really cool component you did here 🔥 

In my phone the animation feels a bit heavy, is not very smooth, perhaps we could imrpove something to speed up the performance?

Also, in android, im used to refreshers that stay put in place and spinning while they are loading, maybe something we could add?

Left some comments about the sync action itself 🙏🏼

PR #4214, opened 2022-08-12
refactor: improve activity initialization

MarkNerdi:
## Summary

Add a preprocessing step, where we collect relevant data from either output or transaction and bundle it into one class. This class is then used for activity creation.
This `processedOutput` class allows us following things in the future:
- linking outputs for claiming, storage deposit return, foundries
- we can determine the resulting activity type in advance, with which we can implement different `Activity` classes for semantically different activities eg. Transactions, Alias, Foundries

## Changelog

```
```

## Relevant Issues

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/profile-manager/api/events/newOutput.ts
```diff
@@ -7,23 +7,24 @@ import { Bech32Helper } from '@lib/bech32Helper'
...
-        account?.depositAddress === address &&
-        !event?.output?.remainder
-    ) {
+    if (output?.address?.type === ADDRESS_TYPE_ED25519 && account?.depositAddress === address && !output?.remainder) {
         syncBalance(account.id)
-        addActivityToAccountActivitiesInAllAccountActivities(
-            account.id,
-            await new Activity().setFromOutputData(event?.output, account, event?.transactionInputs)
+
+        const processedOutput = preprocessGroupedOutputs(
```
nicole-obrien, line 23: Can we assume that this output recieved is not grouped with another existing one?

packages/shared/components/modals/AssetActionsMenu.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="typescript">
     import { localize } from '@core/i18n'
     import { hideAsset, IAsset, unhideAsset, unverifyAsset, VerificationStatus, verifyAsset } from '@core/wallet'
-    import { hideActivitiesForHiddenAssets } from '@core/wallet/actions/hideActivitiesForHiddenAssets'
```
nicole-obrien, line 4: Can import this from @core/wallet

packages/shared/lib/core/wallet/actions/activities/generateAndStoreActivitiesForAccount.ts
```diff
@@ -0,0 +1,23 @@
...
+    // Step 1: process account transactions and outputs into processed transactions
+    const processedTransactions = [...preprocessAccountTransactions(account), ...preprocessAccountOutputs(account)]
+
+    // Step 2: generate activities from processed transactions
+    const activities = generateActivitiesFromProcessedTransactions(processedTransactions, account)
+
+    // Step 3: set account activities with generated activities
+    setAccountActivitiesInAllAccountActivities(account.id, activities)
+
+    hideActivitiesForFoundries(account)
```
nicole-obrien, line 20: Now we group outputs together do we need to hide activities for foundries because only one would have been created? And in the future we might want to create more than one activity out of the processedTransaction, such that an alias was also created.

packages/shared/lib/core/wallet/actions/activities/preprocessAccountOutputs.ts
```diff
@@ -0,0 +1,32 @@
+import { IAccountState } from '@core/account'
+import { preprocessGroupedOutputs } from '@core/wallet/utils/outputs/preprocessGroupedOutputs'
+import { OutputData } from '@iota/wallet'
+import { IProcessedTransaction } from '../../interfaces/processed-transaction.interface'
+
+export function preprocessAccountOutputs(account: IAccountState): IProcessedTransaction[] {
```
nicole-obrien, line 6: I think a better name would be preprocessOutputsForAccount

packages/shared/lib/core/wallet/actions/activities/preprocessAccountTransactions.ts
```diff
@@ -0,0 +1,12 @@
+import { IAccountState } from '@core/account'
+import { preprocessTransaction } from '../../utils'
+import { IProcessedTransaction } from '../../interfaces/processed-transaction.interface'
+
+export function preprocessAccountTransactions(account: IAccountState): IProcessedTransaction[] {
```
nicole-obrien, line 5: I think a better name would be preprocessTransactionsForAccount

nicole-obrien (approved):
I think this is good to merge, there is still lots of improvements I know we are planning to make so I commented some things we might want to do

nicole-obrien (changes requested):
After further testing, I noticed that this transaction is leading to an activity with the second basic output used which is the remainder instead of the first https://explorer.shimmer.network/testnet/transaction/0x93ca693c45f7a288cc682137c9d49fc113bb193906f61550b7c9b60519188e43

nicole-obrien (approved):
After further investigation, it looks like this transaction is a limitation of the data available to us when restoring. We need to resolve two things here: one, transaction syncing when we restore; two, hiding these self transactions unless we can guarantee it was supposed to be a self transaction.

PR #4247, opened 2022-08-17
feat: initial nft tab

jeeanribeiro:
## Summary
This draft PR is a tracker of the initial collectibles tab development progress where all its related tasks should be based and merged to


nicole-obrien (approved):
We only need to set the routers and create initial empty tab component in this task. I cleaned up some of the files for you and disabled the feature flag because it shouldn't be enabled until we have MVP for NFTs.

PR #4279, opened 2022-08-21
refactor: split sendForm and sendConfirmation deep link operations

nicole-obrien:
## Summary
We need to refactor deep link slightly before adding support for native tokens
Todo:
- [x] Add deep link simulator in the developer tools tab
- [x] Split send form and send confirmation form deep link
- [x] Add no back button prop to send confirmation form from deep link
- [x] Add disable expiration time changes for send confirmation deep link
- [x] Add disable gift storage deposit changes for send confirmation deep link

## Changelog

```
```

## Relevant Issues

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/components/popups/SendConfirmationPopup.svelte
```diff
@@ -175,9 +182,16 @@
...
+        {#if disableBack}
+            <Button classes="w-full" secondary onClick={onCancel} disabled={$isTransferring}>
+                {localize('actions.cancel')}
+            </Button>
+        {:else}
+            <Button classes="w-full" secondary onClick={onBack} disabled={$isTransferring}>
+                {localize('actions.back')}
+            </Button>
+        {/if}
+
```
maxwellmattryan, line 194: We can shorten this to:
```svelte
<Button classes="w-full" secondary onClick={disableBack ? onCancel : onBack} disabled={$isTransferring}>
    {localize(`actions.${disableBack ? 'cancel' : 'back'}`)}
</Button>
```

packages/shared/lib/core/wallet/interfaces/send-confirmation-parameters.interface.ts
```diff
@@ -0,0 +1,11 @@
+import { IAsset, Subject } from '@core/wallet'
+
+export interface ISendConfirmationParameters {
+    asset?: IAsset
+    amount?: string | undefined
+    unit?: string
+    recipient?: Subject
+    metadata?: string
+    tag?: string
+    giftStorageDeposit?: boolean
```
maxwellmattryan, line 10: Does this not need the two other send parameters for disabling the gifting of the deposit, etc.? Might be related to the above comment about the different between "send operation parameters" and "send confirmation operation parameters"?

maxwellmattryan (commented):
Looking really good, just have a few minor comments :+1:

packages/shared/locales/en.json
```diff
@@ -1029,7 +1029,9 @@
         "unhideToken": "Unhide token",
         "burnToken": "Burn token",
         "faucetRequest": "Get {token} tokens",
-        "refreshTokenMetadata": "Refresh token metadata"
+        "refreshTokenMetadata": "Refresh token metadata",
+        "test": "Test",
+        "testDeepLink": "Test deep Link"
```
maxwellmattryan, line 1034: This should be "Test deep link"

maxwellmattryan (changes requested):

maxwellmattryan:
Some thoughts:
- We may want to disallow the user from deleting the prefix text of the URL, so that the wrong prefix can never be used. Or if the wrong prefix _is_ used then just display a warning notification against
- When deep links are disabled and the dev tries to use this functionality, we should either **1)** just disable / grey the button out in the first place or **2)** display a warning toast saying "Please enable deep links to use this functionality", maybe even with an action on the notification / alert to do that. I think 2 is a friendlier option since it's clearer what the dev needs to do in order to make the tool work.

maxwellmattryan:
As I understand, deep link error handling is supposed to be improved in a near-future PR. The functionality works, and it's seriously amazing that we now have this and don't have to compile the app. Thanks for the new dev tool :pray:

packages/shared/components/popups/SendConfirmationPopup.svelte
nicole-obrien: I think this is less readable, in my version you can clearly see the two different components to be displayed

packages/shared/lib/core/wallet/interfaces/send-confirmation-parameters.interface.ts
nicole-obrien: Yeah added them

nicole-obrien:
> Some thoughts:
> 
> * We may want to disallow the user from deleting the prefix text of the URL, so that the wrong prefix can never be used. Or if the wrong prefix _is_ used then just display a warning notification against
> * When deep links are disabled and the dev tries to use this functionality, we should either **1)** just disable / grey the button out in the first place or **2)** display a warning toast saying "Please enable deep links to use this functionality", maybe even with an action on the notification / alert to do that. I think 2 is a friendlier option since it's clearer what the dev needs to do in order to make the tool work.
> * I used this deep link (missing `s` in `rms`, `firefly-alpha://wallet/sendConfirmation?address=rm1qpgn2aelvx6qgxyvyyuvk5qrj6ps9mv7s5zlx5gygdzlgh7gpg3cwr536uv&amount=100&metadata=128347213&tag=soonaverse`) and IMO it should have displayed an error or warning, but instead it took me to the dashboard without doing or showing anything

Yeah we will improve error handling of deep links in a separate PR.

As for the deep links being enabled or not, this is a poor feature and just wrong. Thats because it doesnt register or deregister the deep link. But just turns of the handling of deep links and sends the user to the settings page, "technically" handling the deep link anyway. I think we should remove it and open a task to register and deregister the protocol with the os.

maxwellmattryan:
> I think we should remove it and open a task to register and deregister the protocol with the os.

This makes sense :+1:

packages/shared/locales/en.json
```diff
@@ -1526,7 +1530,7 @@
     "tabs": {
         "wallet": "Wallet",
         "staking": "Staking",
-        "developerTools": "Developer Tools"
+        "developerTools": "Developer"
```
maxwellmattryan, line 1533: Should we also change `developerTools` to `developer` to match?

maxwellmattryan (changes requested):
One minor comment / question

packages/shared/locales/ur.json
```diff
@@ -1524,7 +1524,7 @@
     "tabs": {
         "wallet": "Wallet",
         "staking": "Staking",
-        "developerTools": "Developer Tools"
+
```
maxwellmattryan, line 1527: This is missing `"developer"`?

packages/shared/locales/vi.json
```diff
@@ -1524,7 +1524,7 @@
     "tabs": {
         "wallet": "Wallet",
         "staking": "Staking",
-        "developerTools": "Developer Tools"
+
```
maxwellmattryan, line 1527: Same as above.

packages/shared/locales/zh-CN.json
```diff
@@ -1524,7 +1524,7 @@
     "tabs": {
         "wallet": "钱包",
         "staking": "质押",
-        "developerTools": "Developer Tools"
+
```
maxwellmattryan, line 1527: Same as above.

packages/shared/locales/zh-TW.json
```diff
@@ -1524,7 +1524,7 @@
     "tabs": {
         "wallet": "錢包",
         "staking": "質押",
-        "developerTools": "Developer Tools"
+
```
maxwellmattryan, line 1527: Same as above.

maxwellmattryan (changes requested):
Some potential mishaps with migrating some locale strings?

maxwellmattryan (approved):
LGTM :fire:

adamunchained:
> Some thoughts:
> 
> * We may want to disallow the user from deleting the prefix text of the URL, so that the wrong prefix can never be used. Or if the wrong prefix _is_ used then just display a warning notification against
> * When deep links are disabled and the dev tries to use this functionality, we should either **1)** just disable / grey the button out in the first place or **2)** display a warning toast saying "Please enable deep links to use this functionality", maybe even with an action on the notification / alert to do that. I think 2 is a friendlier option since it's clearer what the dev needs to do in order to make the tool work.
> * I used this deep link (missing `s` in `rms`, `firefly-alpha://wallet/sendConfirmation?address=rm1qpgn2aelvx6qgxyvyyuvk5qrj6ps9mv7s5zlx5gygdzlgh7gpg3cwr536uv&amount=100&metadata=128347213&tag=soonaverse`) and IMO it should have displayed an error or warning, but instead it took me to the dashboard without doing or showing anything

just curious here, did Soonaverse generate deep-link without "s"?

PR #4221, opened 2022-08-15
feat: implement ledger onboarding flows

laumair:
## Summary

This PR sets up the onboarding flow for ledger profiles

## Relevant Issues

https://github.com/iotaledger/firefly/issues/3948

### Platforms

> Please select any platforms where your changes have been tested.

- __Desktop__
  - [x] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

**How to setup ledger:** https://hackmd.io/@laumair/S1-Uyrc05

## Checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or modified tests that prove my changes work as intended
- [ ] I have verified that new and existing unit tests pass locally with my changes
- [ ] I have verified that my latest changes pass CI workflows for testing and linting
- [ ] I have made corresponding changes to the documentation

Issue #3949, opened 2022-07-20
[Task]: Allow making transfers using ledger


laumair:
### Task description

Based on https://github.com/iotaledger/firefly/issues/3948, allow users to send tokens using ledger device

### Requirements

TBD

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

Issue #3974, opened 2022-07-21
[Task]: Refactor `popup` module


maxwellmattryan:
### Task description

The existing logic for popups needs to be moved to a new `popup` module in `lib/auxiliary/popup`.

### Requirements

- [x] Create `popup` module in `lib/auxiliary/popup`
- [ ] ~~Consider changing popup `type` to `name` since "type" could also refer to something else (also leaves it open for us to make distinct types of popups`)~~
- [ ] ~~Add `PopupType` (if `type` wasn't renamed, otherwise `Popup`) enum so we don't use hard-coded strings everywhere~~
- [x] Add logic and stores from old popup file (`lib/popup.ts`) to new module
- [x] Add types (`lib/typings/popup.ts`) from old popup file to new module
- [x] Update [handbook specification](https://iotaledger.github.io/firefly/specifications/modules/#auxiliary) as necessary
- [ ] ~~(Opt.) Create a specification page in the handbook detailing popups and how they work / how they're used~~

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #4292, opened 2022-08-22
feat: ledger send flow

laumair:
## Summary

To be reviewed after https://github.com/iotaledger/firefly/pull/4221

## Changelog

Allow sending transfers using ledger device. 

## Relevant Issues

https://github.com/iotaledger/firefly/issues/3949

### Platforms

- __Desktop__
  - [x] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/components/popups/EnableLedgerBlindSigningPopup.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="typescript">
+    import { localize } from '@core/i18n'
+    import { Text, Icon } from 'shared/components'
+    import { openPopup } from 'shared/lib/popup'
```
maxwellmattryan, line 4: Not technically necessary, but we can use the `@lib/...` path alias :+1:

packages/shared/features/onboardingFeatures.ts
```diff
@@ -38,7 +38,7 @@ const onboardingFeatures = {
                     enabled: false,
                 },
                 ledgerBackup: {
-                    enabled: false,
+                    enabled: true,
```
maxwellmattryan, line 41: I think we're supposed to leave the IOTA-based feature flags set to false until V2 can support it

packages/shared/lib/core/ledger/interfaces/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './ledger-connection-state.interface'
 export * from './ledger-extended-status.interface'
 export * from './ledger-status.interface'
+export * from './ledger-send-confirmation-props.interface'
```
maxwellmattryan, line 4: Sometimes exporting in alphabetical order breaks compilation, was that also the case here?

maxwellmattryan (changes requested):
Looks fantastic! Just a few minor changes requested :pray:

packages/shared/features/onboardingFeatures.ts
```diff
@@ -119,15 +119,15 @@ const onboardingFeatures = {
...
                 },
             },
         },
     },
     shimmer: {
         enabled: true,
         mainnet: {
-            enabled: false,
+            enabled: true,
```
maxwellmattryan, line 130: This should be set to `false`

maxwellmattryan (changes requested):
Added another change request :sweat_smile:

maxwellmattryan:
Here are some general comments / feedback after testing:

- When I disconnect the device in the send confirmation popup view, it displays an error to the console but nothing else happens. When I re-connect and unlock the Ledger device (and open the Shimmer app), it says in the console "DevTools was disconnected from the page. Once page is reloaded, DevTools will automatically...".
    
    I think maybe showing a warning or error message here would be helpful. The transaction details have to be re-sent to the Ledger so the user will have to re-trigger that somehow :thinking: 

- I triggered the blind signing warning, then rejected the transaction. I don't see the blind signing steps anymore, but when I went to add an expiry time again it did this:
    
    Screenshot from 2022-08-26 08-40-25

    It would be nice to somehow persist the blind signing instructions popup since the Legder device itself doesn't let you do it directly from that point in the flow, you have to reject the TX first which causes the blind signing instructions to go away.

packages/shared/lib/core/profile-manager/api/events/transactionProgress.ts
```diff
@@ -1,13 +1,67 @@
...
+                              props: {
+                                  mintNativeTokenPopupProps: _mintNativeTokenProps,
+                              },
+                          }),
+                })
+            }
+
+            if (payload === 'PerformingPow') {
+                // TODO: Should we double check the type of popup?
+                closePopup()
```
maxwellmattryan, line 51: We could write a slightly more specific function like `closeLedgerConfirmationPopup`?

maxwellmattryan (changes requested):
Just a few more comments ~

packages/shared/lib/auxiliary/popup/enums/popup-type.enum.ts
```diff
@@ -0,0 +1,3 @@
+export enum PopupType {
+    EnableLedgerBlindSigning = 'enableLedgerBlindSigning',
+}
```
Tuditi, line 3: I like this approach, but we have to do this either for all popups or for none, because now this is the only one that follows this approach.

I would suggest cutting this code and pasting it to a new branch, since it would slow down this PR and create an issue for it under Software Quality milestone.

packages/shared/lib/core/ledger/actions/pollLedgerDeviceStatus.ts
```diff
@@ -1,4 +1,5 @@
 import { get } from 'svelte/store'
+
```
Tuditi, line 2: Could we get rid of the warnings on l17 and l18 as well, please?

packages/shared/lib/core/profile-manager/actions/unsubscribeFromWalletApiEvents.ts
```diff
@@ -0,0 +1,7 @@
+import { get } from 'svelte/store'
+
+import { profileManager } from '../stores'
+
+export function unsubscribeFromWalletApiEvents(): void {
+    get(profileManager)?.clearListeners([])
```
Tuditi, line 6: Shouldn't this be:

```suggestion
    get(profileManager)?.clearListeners(['*'])
```

Otherwise you're not clearing any listeners.

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -0,0 +1,83 @@
...
+            handleTransactionProgressInternal(payload)
+        }
+    } else {
+        console.warn('Transaction progress handler unimplemented: ', payload)
+    }
+}
+
+function handleTransactionProgressInternal(
+    payload: TransactionProgressPayload,
+    isDuringOnboarding: boolean = false
```
Tuditi, line 21: Doesn't make sense to type if you initialize it with a `false`. TS knows infers that it's a boolean

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        const isBlindSigningTransaction = typeof payload !== 'string' && 'PreparedTransactionEssenceHash' in payload
+        const toAddress = isBlindSigningTransaction
+            ? null
+            : sendConfirmationProps.recipient.type === 'address'
+            ? sendConfirmationProps.recipient.address
+            : sendConfirmationProps.recipient.account.depositAddress
+        const toAmount = isBlindSigningTransaction
+            ? null
+            : `${sendConfirmationProps.amount} ${sendConfirmationProps.unit}`
```
Tuditi, line 37: This style of coding is very hard to follow/debug. The following style makes it more clear:
I would extract functions here, since this is only relevant information if you want to how the `toAddress` is derived, but not if you want to understand how `handleTransactionProgressInternal` is working.

```
function getToAddress(subject: Subject): string {
  if (isBlindSigningTransaction) {
    return null
  }
  if (...) {
    return subject.address
  }
  ...
}
```

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -0,0 +1,83 @@
...
+                          needsToShowPopupAfterwards: !isDuringOnboarding,
+                          sendConfirmationPopupProps: sendConfirmationProps,
+                      },
+                  }
+                : {
+                      props: {
+                          mintNativeTokenConfirmationPopupProps: mintNativeTokenConfirmationProps,
+                      },
+                  }),
+        })
```
Tuditi, line 59: Idem as above.

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -0,0 +1,83 @@
...
+        const isBlindSigningTransaction = typeof payload !== 'string' && 'PreparedTransactionEssenceHash' in payload
+        const toAddress = isBlindSigningTransaction
+            ? null
+            : sendConfirmationProps.recipient.type === 'address'
+            ? sendConfirmationProps.recipient.address
+            : sendConfirmationProps.recipient.account.depositAddress
+        const toAmount = isBlindSigningTransaction
+            ? null
+            : `${sendConfirmationProps.amount} ${sendConfirmationProps.unit}`
+        const hash = isBlindSigningTransaction ? payload?.['PreparedTransactionEssenceHash'] : null
```
Tuditi, line 38: This is an exception to the rule, since I can see in one glance what's happening, but don't have to follow a whole branch of `?` `:`. Maybe a coding convention should be put in place that this is only allowed for one-liners.

packages/shared/lib/core/utils/format.utils.ts
```diff
@@ -0,0 +1,22 @@
+import { HEXADECIMAL_PREFIX, HEXADECIMAL_REGEXP } from './constants'
+
+export function formatHexString(
+    hex: string,
+    shouldBeUppercase: boolean = true,
+    shouldIncludePrefix: boolean = true
```
Tuditi, line 6: Same as above, no need to type booleans if you give the default boolean values. TS is smart enough to know what's happening.

packages/shared/lib/core/utils/format.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+): string {
+    if (!hex || typeof hex !== 'string' || !HEXADECIMAL_REGEXP.test(hex)) {
```
Tuditi, line 8: No need for `typeof hex !== 'string'`. You can expect a string here, since you type it as an argument here. If you supply something else, TSC will scream at you!

packages/shared/lib/core/utils/tests/format.test.ts
```diff
@@ -0,0 +1,28 @@
...
+            expect(formatHexString('0x1a2b3c')).toEqual('0x1A2B3C')
+        })
+        it('should handle valid hex strings with modifications', () => {
+            expect(formatHexString('0', true, false)).toEqual('0')
+            expect(formatHexString('0x1', false, false)).toEqual('1')
+            expect(formatHexString('0x1a2b3c', true, true)).toEqual('0x1A2B3C')
+            expect(formatHexString('0x1A2B3C', false, true)).toEqual('0x1a2b3c')
+        })
+    })
+})
```
Tuditi, line 28: Love this :heart: !

Tuditi (changes requested):
I left some comments to address. The app works very smooth already, nice job! There are a couple of things I noticed when testing:

1) I get an error notification (`Unable to use Stronghold with ledger profile`) if I claim a transaction with ledger. However the transaction goes through, since the transaction is claimed.

2) If I reject a transaction, I get a notification popup, but the confirmation modal still shows up. Shouldn't the modals close upon rejecting a tx? (same is true for minting a native token).

3) When claiming a native token the hash of the blind signature is not shown. Rejecting the transaction also results in the modal staying there and there is no way to get rid of it.

packages/shared/components/popups/EnableLedgerBlindSigningPopup.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="typescript">
+    import { onDestroy } from 'svelte'
+    import { Text, Icon } from 'shared/components'
+    import { localize } from '@core/i18n'
+    import { ledgerDeviceStatus } from '@core/ledger'
+    import { closePopup, openPopup } from '@lib/popup'
+
+    export let sendConfirmationPopupProps = null
+    export let mintNativeTokenPopupProps = null
```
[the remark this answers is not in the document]
maxwellmattryan: > I always feel like initializing something as null doesn't really make sense, since it could just as well be undefined. Maybe we should discuss this and create a coding convention out of it. Because sometime we initialize it as undefined or as null.

In my mind, I thought of `null` as a way of explicitly saying we're initializing this as an empty variable. It could be `undefined` but since that value / type is returned from most functions rather than `null`, using both keeps the intention and code a bit clearer IMO.

packages/shared/lib/auxiliary/popup/enums/popup-type.enum.ts
maxwellmattryan: My thinking was that we can do it incrementally, so if your changes touch a popup that isn't being used this way then you should try and change it. We don't have to have it for all or none - it's backwards compatible so long as we use the same string value for each popup type. I also think rather than `type`, `name` is a better descriptor, but that can be in a separate PR / task. 

FYI there already exists a backlog [task](https://github.com/iotaledger/firefly/issues/3974) to refactor popups :+1:

packages/shared/components/popups/EnableLedgerBlindSigningPopup.svelte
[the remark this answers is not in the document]
maxwellmattryan: After thinking about it, `undefined` does make more sense idiomatic to the language since declared but uninitialized values are `undefined`. But rather than initializing the variable as `undefined` explicitly, we just declare the object with a type and it will automatically be `undefined`. It's also slightly less code too.

packages/shared/lib/core/profile-manager/actions/unsubscribeFromWalletApiEvents.ts
maxwellmattryan: I haven't looked under the hood to confirm if the listeners are actually destroyed, but I was just going off of this NodeJS [example](https://github.com/iotaledger/wallet.rs/blob/develop/bindings/nodejs/examples/7-events.js#L61) from the `wallet.rs` repo.

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
maxwellmattryan: Fair enough, but for more complicated parameters like a profile manager, it is also inferred by TS but without the type there it's not dev friendly. Although not the same as a `boolean`, the inconsistency bothers me :sweat_smile:

packages/shared/components/popups/LedgerTransaction.svelte
```diff
@@ -1,105 +1,68 @@
...
+    import { resetLedgerSendConfirmationProps, resetLedgerMintNativeTokenConfirmationProps } from '@core/ledger'
+    import { formatHexString } from '@core/utils'
+    import { openPopup } from '@lib/popup'
 
-    export let locale: Locale
-
-    export let remainderAddress = ''
-    export let remainderAmount = null
     export let toAddress = ''
     export let toAmount = null
```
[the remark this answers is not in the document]
maxwellmattryan: See this [comment](https://github.com/iotaledger/firefly/pull/4292#discussion_r959813226).

maxwellmattryan:
@Tuditi I've made adjustments accordingly - I've decided to leave minting and async tx claiming for other tasks to help speed up the process. I've also disabled Ledger features for onboarding since the profile isn't technically feature-complete.

packages/shared/components/popups/LedgerTransactionPopup.svelte
```diff
@@ -0,0 +1,69 @@
...
+        ledgerMintNativeTokenConfirmationProps,
+    } from '@core/ledger'
+    import { formatHexString } from '@core/utils'
+    import { openPopup } from '@lib/popup'
+
+    export let toAddress: string
+    export let toAmount: string
+    export let hash: string
+
+    const hasSendConfirmationProps = (toAddress && toAmount) || hash
```
maxwellmattryan, line 18: We'll have to adjust this logic for native token minting, but it should be done in the corresponding PR.

docs/guides/coding-conventions/components.md
```diff
@@ -28,6 +28,29 @@ The organization of a Svelte component **must** start with the script, followed
...
+```
+
+**Good**
+
+```typescript
+export let myPropObject: IPropObject
+export let showSomething: boolean
+```
+
 ## Reactivity
```
Tuditi, line 54: Excellent!

packages/shared/lib/core/ledger/utils/openLedgerConfirmationPopup.ts
```diff
@@ -0,0 +1,37 @@
...
+        let props: PopupProps
+        switch (true) {
+            case isPreparedTransaction(payload):
+                props = {
+                    ...deconstructLedgerSendConfirmationProps(),
+                    needsToShowPopupAfterwards: !isDuringOnboarding,
+                }
+                break
+            case isPreparedTransactionEssenceHash(payload):
+                props = { hash: payload?.['PreparedTransactionEssenceHash'] }
```
Tuditi, line 24: Here optional chaining is not required, since you already check whether payload contains the `PreparedTransactionEssenceHash` in l.23

docs/guides/coding-conventions/components.md
Tuditi: image

packages/shared/lib/core/ledger/interfaces/ledger-mint-native-token-confirmation-props.store.ts
```diff
@@ -1,4 +1,4 @@
-export interface LedgerMintNativeTokenConfirmationProps {
+export interface ILedgerMintNativeTokenConfirmationProps {
```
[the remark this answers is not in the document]
maxwellmattryan: I am not seeing any errors in this popup wrt to this type. Also, it's in our conventions to use interfaces whenever defining the structure of an object.

packages/shared/lib/core/ledger/interfaces/ledger-send-confirmation-props.interface.ts
```diff
@@ -1,6 +1,6 @@
 import { IAsset, Subject } from '@core/wallet'
 
-export interface LedgerSendConfirmationProps {
+export interface ILedgerSendConfirmationProps {
```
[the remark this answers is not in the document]
maxwellmattryan: See above comment.

packages/shared/lib/core/ledger/utils/openLedgerConfirmationPopup.ts
```diff
@@ -0,0 +1,37 @@
+import { PopupProps } from '@auxiliary/popup'
+import { isPreparedTransaction, isPreparedTransactionEssenceHash } from '@core/profile-manager'
+import { TransactionProgressEventPayload } from '@core/profile-manager/api/types'
+import { openPopup } from '@lib/popup'
+
+import { deconstructLedgerSendConfirmationProps } from '../helpers'
+
+export function openLedgerConfirmationPopup(
+    payload?: TransactionProgressEventPayload,
```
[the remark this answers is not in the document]
maxwellmattryan: It's for a guard to check that the payload is really there. We've shot ourselves in the foot many times passing `undefined` values to functions where it expected something and because we didn't check that the object was really there Firefly would crash.

packages/shared/lib/core/profile-manager/api/types/transaction-progress-payload.type.ts
```diff
@@ -1,8 +1,8 @@
...
+import { PreparedTransactionEssenceHashType } from './prepared-transaction-essence-hash.type'
 
-export type TransactionProgressPayload =
+export type TransactionProgressEventPayload =
     | 'PerformingPow'
     | 'SigningTransaction'
-    | RegularTransaction
-    | BlindLedgerTransaction
+    | PreparedTransaction
+    | PreparedTransactionEssenceHashType
```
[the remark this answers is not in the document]
maxwellmattryan: Was a mistake, will fix :+1:

packages/shared/lib/core/profile-manager/api/types/index.ts
```diff
@@ -1,5 +1,5 @@
-export * from './blind-ledger-transaction.type'
 export * from './new-output-event.type'
-export * from './regular-transaction.type'
+export * from './prepared-transaction.type'
+export * from './prepared-transaction-essence-hash.type'
 export * from './transaction-inclusion-event.type'
 export * from './transaction-progress-payload.type'
```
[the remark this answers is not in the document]
maxwellmattryan: Not currently seeing any type errors in `openLedgerConfirmationPopup` but will export :+1:

packages/shared/lib/core/ledger/utils/openLedgerConfirmationPopup.ts
maxwellmattryan: True, just an automatic habit for me :sweat_smile:

docs/guides/coding-conventions/components.md
maxwellmattryan: Removed it :sweat_smile:

packages/shared/lib/core/ledger/utils/openLedgerConfirmationPopup.ts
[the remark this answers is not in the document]
Tuditi: But this doesn't solve the issue, now it just fails silently. Maybe we should throw an error if there is no payload?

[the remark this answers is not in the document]
maxwellmattryan: Yeah that wouldn't be a bad idea, I'll add really quick

Tuditi (approved):
Looks good and works excellent!

maxwellmattryan (approved):

PR #4308, opened 2022-08-23
refactor: activity payload types

MarkNerdi:
## Summary

...

## Changelog

```
```

## Relevant Issues

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/wallet/classes/activity.class.ts
```diff
@@ -96,106 +47,69 @@ export class Activity implements IActivity {
...
-                : getAmountFromOutput(output) - this.storageDeposit
+            this.data = getFoundryActivityData(processedTransaction)
         }
     }
 
     updateFromPartialActivity(partialActivity: Partial<IActivity>): void {
         Object.assign(this, partialActivity)
     }
 
+    updateDataFromPartialActivity(partialData: Partial<TransactionActivityData | FoundryActivityData>): void {
```
MarkNerdi, line 63: for better dev-ex, i'd like to have the `type` field required in this parameter, do you have any suggestions how to do that?

nicole-obrien (dismissed):

packages/shared/lib/core/wallet/classes/activity.class.ts
```diff
@@ -96,153 +43,60 @@ export class Activity implements IActivity {
...
     }
 
+    updateDataFromPartialActivity(
+        partialData: IPartialTransactionActivityDataWithType | IPartialFoundryActivityDataWithType
+    ): void {
+        if (partialData.type === ActivityType.Transaction && this.data.type === ActivityType.Transaction) {
+            this.data = { ...this.data, ...partialData }
+        } else if (partialData.type === ActivityType.Foundry && this.data.type === ActivityType.Foundry) {
+            this.data = { ...this.data, ...partialData }
+        }
```
Tuditi, line 66: I'm not sure whether the conditionals are necessary here. Both branches do the same thing, right?

packages/shared/lib/core/wallet/utils/isFilteredActivity.ts
```diff
@@ -78,25 +80,39 @@ export function isFilteredActivity(activity: Activity): boolean {
...
+            activity.data.type === ActivityType.Transaction &&
+            activity.data.direction !== ActivityDirection.In
+        ) {
             return true
         }
-        if (filter.type.selected === TypeFilterOption.Outgoing && activity.direction !== ActivityDirection.Out) {
+        if (
+            filter.type.selected === TypeFilterOption.Outgoing &&
+            activity.data.type === ActivityType.Transaction &&
+            activity.data.direction !== ActivityDirection.Out
```
Tuditi, line 107: Maybe it starts to make sense to add a helper function `isOutgoingTransaction()` and `isIncomingTransaction`? Also again it's clearer and less error prone to use `===` than `!==`.

MarkNerdi: regarding the `===` and `!==`: i generally agree, but i think in this case `!==` makes more sense. In the filter i have the selected `FilterOption.Outgoing` therefore i want to exclude all which are not `DIrection.Out`

packages/shared/lib/core/wallet/classes/activity.class.ts
MarkNerdi: initially i had the check:
`if (partialData.type === this.data.type`  to assure that we don't set foundry data on a transaction or vice versa. But then typescript is angry because we have `data` and `partialData` which both can be `TypeA | TypeB`. With this implementation we narrow down both of them and assure typescript, that `TypeA` is only assigned to `TypeA`

packages/shared/lib/core/wallet/utils/isFilteredActivity.ts
Tuditi: But aren't all outgoing transactions ones with `Direction.out`? If you selected outgoing don't you want to have the outgoing transactions from all the transactions? Maybe I'm missing something... :laughing:

Tuditi (approved):
LGTM!

PR #4320, opened 2022-08-23
enhancement: add error handling during login

### LINKED ISSUE #3918: [Task]: Handle error for incorrect device time
### Task description

Some users are suffering from a bug that prevents their profiles from accessing the wallet view. There application hangs on the Firefly loading screen, and they continuously see the Firefly logo's loading animation. Restarting the application and logging into the profile also produces the same issue.

OS's reporting this error:
- Windows 10
- Windows 11

I am able to reproduce this behavior on my Windows 11 laptop.

### Requirements

Capture error that causes this freeze by setting up a new specialized Firefly alpha/beta version that enables developer console.

### Creation checklist

### LINKED ISSUE #4123: [Task]: Add error handling during login
### Task description

We should improve error handling during login. And if the user can not login they should be logged out and explained why. IF a profile is corrupt or unrecoverable we could even give them an opportunity to delete it from the login menu.

### Requirements

TBD

### Creation checklist

- [x] I have added the most appropriate labels
- [x] I have linked the correct milestone and/or project

nicole-obrien:
## Summary

Improve core error module to include handling errors and use the new functionality to add error handling to the login function.

## Changelog

```
- Make sure errors are bubbling up to login() catch block
- Create handleError functions in core/error/utils
- Add constants and enums to core/error to support errorHandling
```

## Relevant Issues

Closes #4123 and Closes #3918

### Platforms

- __Desktop__
  - [x] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/error/constants/api-error-parameters.constant.ts
```diff
@@ -0,0 +1,21 @@
+import { ClientError, ErrorFromApi } from '../enums'
+import { IErrorParameters } from '../interfaces'
+
+export const API_ERROR_PARAMETERS: Readonly<
+    { [key in ErrorFromApi]?: { [key in ClientError]?: Partial<IErrorParameters> } }
```
MarkNerdi, line 5: i think this type isn't correct.
If the toplevel key is `ErrorFromApi.ClientError` then the subkeys can be of type `ClientError`.
But if the toplevel key is e.g. `ErrorFromApi.InsufficientFundsError`  (which you get from walletrs as `{type: "InsufficientFunds", message: "insufficient amount of funds ..."}`, we wouldn't add these subtypes in the `ClientError` enum

packages/shared/lib/core/error/utils/handleClientError.ts
```diff
@@ -0,0 +1,35 @@
+import { API_ERROR_PARAMETERS, CLIENT_ERROR_REGEXES } from '../constants'
+import { ClientError } from '../enums'
+import { IError } from '../interfaces'
+import { handleError } from './handleError'
+import { handleGenericError } from './handleGenericError'
+
+export function handleClientError(error: IError): void {
```
MarkNerdi, line 7: i would generalize this. i think we would do the exact same for not only errors with `type: ClientError` right?

MarkNerdi (changes requested):
In general it's a nice way to handle error

But the namings are only adapted to Clienterrors, even though we would do the same things for non-`ClientErrors` thrown by wallet.rs. It wouldn't really make sense to create different Enums and Regexs for different `error.type`s. 

Therefore I think we can remove the layer `ErrorFromApi`, because what interests us essentially is only the error message we get.

So 
CLIENT_ERROR_REGEX -> ERROR_REGEX 
ClientError -> (Api?)Error

wdyt?

nicole-obrien:
@MarkNerdi there are a whole load of error types that can be thrown by wallet.rs. Some of them we will just have a catch all error for some we want to check the error message to see if it's a specific error we want to handle in a certain way. Because the switch statements for the regexes would potentially evaluate them all I think it's more efficient and sensible to only do the regexes for the types and messages we want. Rather than creating one linear list. Hence it's probably more efficient to filter on the type first. That's also because the fall back handling for each error type might be different. We might want to show error notification for client errors but not another type etc

MarkNerdi (approved):

PR #4337, opened 2022-08-25
refactor: activity hiding step as preprocessing

### LINKED ISSUE #4255: [Task] Refactor hiding step for activities
Instead of creating activities for each processed transaction and then hiding them, we should execute this earlier, such that unwanted processed transactions get deleted and only the wanted activities get created

MarkNerdi:
## Summary

For better performance and clear understanding, we now decide which activities we want to show in the  ActivityList before we generate them.

## Changelog

```
- link processed transactions with claiming transactions
- don't generate activities from transactions which are claiming transactions 
```

## Relevant Issues

Closes #4255 

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/wallet/actions/activities/linkTransactionsWithClaimingTransactions.ts
```diff
@@ -0,0 +1,67 @@
...
+export function linkTransactionsWithClaimingTransactions(
+    transactions: IProcessedTransaction[],
+    account: IAccountState
+): IProcessedTransaction[] {
+    const resultingTransactions = []
+    const transactionsIncludedAsClaimingTransactions = []
+
+    const claimedAccountActivities = get(claimedActivities)?.[account.id]
+
+    for (const transaction of transactions.sort((t1, t2) => (t1.time > t2.time ? 1 : -1))) {
```
nicole-obrien, line 16: Can we sort into a separate variable and then iterate on that, so its easier to read.

packages/shared/lib/core/wallet/actions/activities/linkTransactionsWithClaimingTransactions.ts
```diff
@@ -0,0 +1,67 @@
...
+        if (transactionsIncludedAsClaimingTransactions.includes(transaction.transactionId)) {
+            continue
+        } else if (transaction.outputs.some((_output) => !isOutputAsync(_output))) {
+            resultingTransactions.push(transaction)
```
nicole-obrien, line 20: Currently we are looping through all transactions at the top level, and then if we find an async transaction then we are looping through all transactions again.

I think it would be more efficient to do the linking in this conditional. So you loop through the maximum list only once, and then only loop through transactions added to the async transactions.

Three conditionals being:
1. Do we already have cached data if so then link and update lists straight away
2. Is the transaction async? If so add it to the async transactions list but it is still waiting to be linked.
3. Id transaction is not async then we can check the list of async transactions to see if it is matches this one.

We could also optomise if we made the assumption that an async transaction is only going to a many to 1 link with transactions because we can remove it from the async list after its linked so we don't check it again.

nicole-obrien (changes requested):
Looks good, but I think we can improve the alogorithm as discussed on the call. And left a comment directly.

nicole-obrien (changes requested):
Nice change to the algorithm, however there is currently a flaw.
We are not checking the direction of the async transactions so this logic is also happening to outgoing async transactions, meaning if they expire and we use the outputs as inputs for another transaction then we are incorrectly linking to transactions.

Also lets add some comments because its not trivial to read.

nicole-obrien (approved):
I think this works for now, but I suspect we need to refactor this again. Because a claiming is per output or activity and in this we are linking one claim per transaction where a transaction could have multiple async outputs.

Issue #4411, opened 2022-09-02
[Task]: Resolve closing UX when the use is prompted to enable blind signing


nicole-obrien:
### Task description

If a transaction requires blind signing and it is not enabled on the ledger then we display a popup for this when they go to send a transaction. Once the user dismisses the message on the device they are able to close the popup, but they are not made aware of this as it is just the background click that works.

We should update the ux so its more clear and easy to use. I think we should add the close button at the top when they are able to close the popup, or we could even have a button at the bottom which is disabled until they are in a position to close the popup. We could even expand that to have a back button to allow them to go back in the flow.

### Requirements

- [ ] TBD

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

Issue #4410, opened 2022-09-02
[Task]: Continue send flow after user enables blind signing on ledger


nicole-obrien:
### Task description

If a transaction requires blind signing and it is not enabled on the ledger then we display a popup for this when they go to send a transaction. Once the user enables blind signing the popup closes and they have to start the send flow again. We should change this so that when the user enables blind signing it continues the flow where they are prompted to verify the transaction.

### Requirements

- [ ] Verify the transaction after user enables blind signing in the send flow

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

Issue #4291, opened 2022-08-22
[Task] While sending is pending, the SendForm shows no expiration date


MarkNerdi:
When I create a transaction with expiration date and confirm it, the SendForm while loading shows that the expiration date is `none`

jeeanribeiro:
Apparently the pull request that would close this issue was merged

PR #4423, opened 2022-09-02
chore: polish send flow popup state

Tuditi:
## Summary

Polished send flow for ledger and stronghold. Fixed that sendConfirmation popup was reset when blind signing had to be enabled or stronghold had to be unlocked.

...

## Changelog

```
- replaced send-form-parameters with new-transaction-details
- removed props for popups and used stores instead
- added expiration time enum
- replaced unnecessary reactive statements

```

## Relevant Issues

closes: #4291
closes: #4410
closes: #4411 

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/components/atoms/buttons/SendButton.svelte
```diff
@@ -1,9 +1,11 @@
 <script lang="typescript">
     import { localize } from '@core/i18n'
+    import { resetNewTransactionDetails } from '@core/wallet'
     import { openPopup } from '@lib/popup'
     import { OnboardingButton } from 'shared/components/atoms'
 
-    function handleSendClick() {
+    function handleSendClick(): void {
```
maxwellmattryan, line 7: Can you rename this (and similarly any other click handlers) to `onSendClick`. Not in the conventions, but we use `onXYZ` for some component props, just helps to make it cleaner.

packages/shared/components/modals/ExpirationTimePickerMenu.svelte
```diff
@@ -8,7 +14,7 @@
     import { fade } from 'svelte/transition'
 
     export let value: Date
-    export let selected: ExpirationTimeSelection = 'none'
+    export let selected = ExpirationTime.None
```
maxwellmattryan, line 17: Can you add the type back since this is a prop?

packages/shared/components/modals/ExpirationTimePickerMenu.svelte
```diff
@@ -1,5 +1,11 @@
 <script lang="typescript" context="module">
-    export type ExpirationTimeSelection = 'none' | '1hour' | '1day' | '1week' | 'custom'
+    export enum ExpirationTime {
+        None,
+        OneHour,
+        OneDay,
+        OneWeek,
+        Custom,
```
maxwellmattryan, line 7: Do we need the enum variant values too? Doesn't look like they're used for the locales, but just making sure.

packages/shared/components/popups/TokenInformationPopup.svelte
```diff
@@ -23,7 +23,7 @@
         }
     }
 
-    function handleVerify() {
+    function handleVerify(): void {
```
maxwellmattryan, line 26: Should be `onVerifyClick`

packages/shared/components/popups/TokenInformationPopup.svelte
```diff
@@ -37,13 +37,11 @@
         }
     }
 
-    function handleSend() {
+    function handleSend(): void {
```
maxwellmattryan, line 40: Should be `onSendClick`

packages/shared/lib/core/wallet/actions/sendOutput.ts
```diff
@@ -15,6 +15,7 @@ export async function sendOutput(output: OutputTypes): Promise<void> {
         isTransferring.set(true)
         const account = get(selectedAccount)
         const transaction = await account.sendOutputs([output], DEFAULT_TRANSACTION_OPTIONS)
+        resetNewTransactionDetails()
```
maxwellmattryan, line 18: Why do we reset here? Maybe add an explanatory `NOTE: ...` comment or something.

packages/shared/lib/core/wallet/interfaces/index.ts
```diff
@@ -14,6 +14,6 @@ export * from './irc30-metadata.interface'
 export * from './persisted-asset.interface'
 export * from './persisted-assets.interface'
 export * from './send-confirmation-parameters.interface'
-export * from './send-form-parameters.interface'
+export * from './new-transaction-details.interface'
```
maxwellmattryan, line 17: Does alphabetical order work here?

packages/shared/lib/core/wallet/interfaces/new-transaction-details.interface.ts
```diff
@@ -1,10 +1,13 @@
...
 
-export interface ISendFormParameters {
+export interface INewTransactionDetails {
+    amount?: string
     asset?: IAsset
-    amount?: string | undefined
-    unit?: string
-    recipient?: Subject
+    expirationDate?: Date
+    // TODO: do we need this?
```
maxwellmattryan, line 7: :eyes: Any corresponding task for this?

packages/shared/lib/core/wallet/stores/new-transaction-details.store.ts
```diff
@@ -0,0 +1,16 @@
+import { writable, Writable } from 'svelte/store'
+import { INewTransactionDetails } from '..'
+
+export const newTransactionDetails: Writable<INewTransactionDetails> = writable({})
```
maxwellmattryan, line 4: We typically initialize to `null` - we have to be careful initializing as an empty object since those are truthy whereas `null` is falsy.

packages/shared/lib/core/wallet/stores/index.ts
```diff
@@ -4,4 +4,4 @@ export * from './hidden-activities.store'
 export * from './persisted-assets.store'
 export * from './selected-account-activities.store'
 export * from './selected-account-assets.store'
-export * from './send-form-parameters.store'
+export * from './new-transaction-details.store'
```
maxwellmattryan, line 7: Can be in alphabetical order?

packages/shared/components/popups/TokenInformationPopup.svelte
```diff
@@ -9,7 +9,7 @@
     export let asset: IAsset
     export let activityId: string
 
-    function handleSkip() {
+    function handleSkip(): void {
```
maxwellmattryan, line 12: Rename accordingly if is directly called as a result of a user action

packages/shared/components/molecules/ExpirationTimePicker.svelte
```diff
@@ -1,14 +1,15 @@
 <script lang="typescript">
     import { Text, ExpirationTimePickerMenu, Icon } from 'shared/components'
     import { formatDate, localize } from '@core/i18n'
+    import { ExpirationTime } from 'components/modals/ExpirationTimePickerMenu.svelte'
 
     export let value: Date
-    export let initialSelected: 'none' | '1hour' | '1day' | '1week' = 'none'
-    export let disabled
+    export let initialSelected = ExpirationTime.None
+    export let disabled = false
```
maxwellmattryan, line 8: Let's add types since these are props

packages/shared/components/modals/ExpirationTimePickerMenu.svelte
```diff
@@ -37,27 +43,26 @@
...
+            case ExpirationTime.OneWeek:
                 value = dateIn1Week
                 break
-            case 'custom':
+            case ExpirationTime.None:
+                value = null
                 break
+            case ExpirationTime.Custom:
             default:
-                value = null
```
maxwellmattryan, line 55: Should we not have the `value = null` statement for the `default` case?

maxwellmattryan (changes requested):
Have some minor changes requested - generally looks good though :+1:

nicole-obrien (commented):
Does this work with the deep links? SendForm and SendConfirmation? I only see changes to SendForm

packages/shared/components/modals/ExpirationTimePickerMenu.svelte
MarkNerdi: In theory we never have the `default` case. But if we have it, its better to keep the current value instead of resetting it

packages/shared/components/modals/ExpirationTimePickerMenu.svelte
MarkNerdi: no we dont need them

packages/shared/lib/core/wallet/stores/new-transaction-details.store.ts
Tuditi: Initializing this to `null` throws errors when destructuring the object after it has been reset or `send Funds` button is clicked. Either we give it sensible default values or create an empty object and fill it in later and never expect this to be null. We never evaluate this object to be truthy or not.

packages/shared/components/popups/SendConfirmationPopup.svelte
```diff
@@ -80,6 +74,18 @@
         type: ActivityType.Transaction,
     }
 
+    $: initialExpirationDate = getInitialExpirationDate()
```
maxwellmattryan, line 77: Not sure if this needs to be reactive since it's just an initialization. I think you'll have to add a date function parameter then pass `expirationDate` to be reactive to that variable, iiuc. Should this just be a regular variable?

packages/shared/components/popups/SendConfirmationPopup.svelte
```diff
@@ -99,19 +105,9 @@
...
-            setLedgerSendConfirmationProps({
-                asset,
-                amount,
-                unit,
-                recipient,
-                internal: false,
-                metadata,
-                tag,
-            })
+        if (!get(isSoftwareProfile)) {
```
maxwellmattryan, line 108: This should probably be `if ($isActiveLedgerProfile || $isOnboardingLedgerProfile) { ... }`. You probably don't need the onboarding profile since the only sending possible in all of the onboarding flows is specifically in the Shimmer claiming flow.

packages/shared/lib/auxiliary/deep-link/errors/amount-not-a-number.error.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseError } from '@core/error'
+
+export class AmountNotANumberError extends BaseError {
+    constructor(amountParam: string) {
+        const message = `Amount in deep link is not a number '${amountParam}'`
+        super({
+            message,
+            showNotification: true,
+            saveToErrorLog: true,
+            logToConsole: true,
```
maxwellmattryan, line 10: You can spread the `DEFAULT_APP_ERROR_PARAMETERS` constant here to save some code, although the only difference is that `logToConsole` is `false` since it's saved to the error log. It's also localized by default, so we should add a new locale entry for this one.

packages/shared/lib/auxiliary/deep-link/errors/invalid-address.error.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseError } from '@core/error'
+
+export class InvalidAddressError extends BaseError {
+    constructor(address: string) {
+        const message = `Address or prefix is not valid for ${address}`
+        super({
+            message,
+            showNotification: true,
+            saveToErrorLog: true,
+            logToConsole: true,
```
maxwellmattryan, line 10: Same as above ^

packages/shared/lib/auxiliary/deep-link/errors/no-address-specified.error.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseError } from '@core/error'
+
+export class NoAddressSpecifiedError extends BaseError {
+    constructor() {
+        const message = 'No address specified in the url path'
+        super({
+            message,
+            showNotification: true,
+            saveToErrorLog: true,
+            logToConsole: true,
```
maxwellmattryan, line 10: Same as above ^

packages/shared/lib/auxiliary/deep-link/handlers/wallet/handleDeepLinkWalletContext.ts
```diff
@@ -28,11 +27,14 @@ export function handleDeepLinkWalletContext(url: URL): void {
         case WalletOperation.SendConfirmation:
             handleDeepLinkSendConfirmationOperation(url.searchParams)
             break
-        default:
+        default: {
+            const message = `Unrecognized wallet operation '${pathnameParts[0]}'`
```
maxwellmattryan, line 31: We should probably add a locale entry for this too?

packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/getAmountFromSearchParam.ts
```diff
@@ -0,0 +1,16 @@
+import { AmountNotANumberError, SendOperationParameter } from '@auxiliary/deep-link'
+import { ITokenMetadata, parseRawAmount } from '@core/wallet'
+
+export function getAmountFromSearchParam(searchParams: URLSearchParams, tokenMetadata: ITokenMetadata): string {
```
maxwellmattryan, line 4: Because this isn't a specific operation handler function, we could probably create a `utils` folder within the `deep-link` module for this? Or a helper.

packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/handleDeepLinkSendConfirmationOperation.ts
```diff
@@ -1,26 +1,30 @@
 import { get } from 'svelte/store'
-import { addError } from '@core/error'
+import { InvalidAddressError, NoAddressSpecifiedError } from '@auxiliary/deep-link'
```
maxwellmattryan, line 2: Any internal module imports from `deep-link` should be with a relative path.

packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/handleDeepLinkSendConfirmationOperation.ts
```diff
@@ -1,26 +1,30 @@
...
+    INewTransactionDetails,
+    Subject,
+    updateNewTransactionDetails,
+    visibleSelectedAccountAssets,
+} from '@core/wallet'
 import { isValidAddressAndPrefix } from '@lib/address'
 import { openPopup } from '@lib/popup'
 
 import { SendOperationParameter } from '../../../enums'
+import { getAmountFromSearchParam } from '@auxiliary/deep-link/handlers/wallet/operations/getAmountFromSearchParam'
```
maxwellmattryan, line 15: Same as above ^

packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/handleDeepLinkSendFormOperation.ts
```diff
@@ -1,22 +1,26 @@
-import { addError } from '@core/error'
+import { InvalidAddressError, NoAddressSpecifiedError } from '@auxiliary/deep-link/errors'
+import { getAmountFromSearchParam } from '@auxiliary/deep-link/handlers/wallet/operations/getAmountFromSearchParam'
```
maxwellmattryan, line 2: The should be relative imports since they're from the same module.

packages/shared/lib/core/ledger/constants/use-ledger-simulator.constant.ts
```diff
@@ -1 +1 @@
-export const USE_LEDGER_SIMULATOR = false
+export const USE_LEDGER_SIMULATOR = true
```
maxwellmattryan, line 1: Probably meant to set this back to `false` :stuck_out_tongue_winking_eye:

packages/shared/lib/core/ledger/stores/ledger-prepared-output.store.ts
```diff
@@ -0,0 +1,4 @@
+import { OutputTypes } from '@iota/types'
+import { writable } from 'svelte/store'
+
+export const ledgerPreparedOutput = writable<OutputTypes>()
```
maxwellmattryan, line 4: I imagine this is initialized as `undefined` rather than `null`, although thinking it best to explicitly pass `null` since we've been doing that elsewhere.

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -10,42 +9,54 @@ import {
...
 import { closePopup, openPopup } from '@lib/popup'
-
 import { TransactionProgressEventPayload } from '../types'
+import { deconstructLedgerVerificationProps } from '@core/ledger/helpers'
 
 export function handleTransactionProgressEvent(accountId: string, payload: TransactionProgressEventPayload): void {
     if (get(isActiveLedgerProfile)) {
         if (get(selectedAccountId) === accountId) {
-            handleTransactionProgressInternal(payload)
+            TransactionProgressEventForLedger(payload)
```
maxwellmattryan, line 18: Should we rename this function to `handleTransactionProgressEventForLedger` since `TransactionProgressEventForLedger` doesn't quite follow the conventions / seems to add a new one?

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
```diff
@@ -10,42 +9,54 @@ import {
...
+                })
             } else {
-                openPopup({ type: 'enableLedgerBlindSigning', hideClose: true, preventClose: true })
+                openPopup({
+                    type: 'enableLedgerBlindSigning',
+                    hideClose: true,
+                    preventClose: true,
+                    props: {
+                        hash: payload?.['PreparedTransactionEssenceHash'],
+                    },
```
maxwellmattryan, line 55: Not sure if we need to pass the hash here since it's the enable blind signing instructions popup.

packages/shared/lib/core/wallet/stores/mint-token-details.store.ts
```diff
@@ -0,0 +1,30 @@
...
+    totalSupply: undefined,
+    circulatingSupply: undefined,
+    decimals: 0,
+    symbol: undefined,
+    description: undefined,
+    url: undefined,
+    logoUrl: undefined,
+})
+
+export function updateMintTokenDetails(payload: IMintTokenDetails): void {
```
maxwellmattryan, line 15: This should either be called `setMintTokenDetails` or have the logic so that it updates including the existing state rather than setting it.

packages/shared/lib/core/wallet/stores/new-transaction-details.store.ts
maxwellmattryan: I think sensible default values make sense then, you could create a new `DEFAULT_NEW_TRANSACTION_DETAILS` constant :+1:

maxwellmattryan (changes requested):
Sorry to be pedantic - have requested some minor changes :pray:

packages/shared/components/popups/MintNativeTokenFormPopup.svelte
```diff
@@ -24,19 +17,23 @@
     } from 'shared/components'
     import { FontWeight } from '../Text.svelte'
     import { onMount } from 'svelte'
+    import { updateMintTokenDetails, mintTokenDetails } from '@core/wallet/stores/mint-token-details.store'
 
-    export let name: string
-    export let totalSupply: number
-    export let circulatingSupply: number
-    export let decimals = 0
```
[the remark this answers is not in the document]
Tuditi: This code has been removed :rofl:

packages/shared/components/popups/SendConfirmationPopup.svelte
Tuditi: You're right, this doesn't need to be reactive

packages/shared/lib/auxiliary/deep-link/errors/amount-not-a-number.error.ts
Tuditi: I think it's more clear to have it explicit what happens here than to have one line of code less. Our error handling is not on point yet, so let's address this in a separate issue. I'll add the locale's however.

packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/getAmountFromSearchParam.ts
Tuditi: What's the difference between a `helper` or `utils` module?

packages/shared/lib/core/profile-manager/api/events/transaction-progress.event.ts
Tuditi: Whoops, I don't know where this comes from :man_shrugging:

packages/shared/lib/core/wallet/stores/new-transaction-details.store.ts
Tuditi: The properties' values depend on what is shown on the `sendFormPopup` initially. All properties are optionally present, since we need compatibility with deep links, but also be able to start a blank transaction. A `DEFAULT_NEW_TRANSACTION_DETAILS` is an empty object. Do you want to make this explicit?

packages/shared/lib/core/wallet/interfaces/new-transaction-details.interface.ts
Tuditi: We cleaned it up :broom:

maxwellmattryan (approved):
LGTM :fire:

PR #4494, opened 2022-09-07
fix: pin code change issues on iOS

### LINKED ISSUE #4473: [Task]: Fix pin component at Change pin code setting
When we press a key the last character gets deleted and readded. This could be related with the latest fixes on pin component.

Image

paul-boegelsack:
## Summary

This PR removes the focus from all pin code fields after the pin has been changed to prevent further entry. This prevents strange input behaviour on iOS.

## Changelog

```
- Adds blur function to pin component
- Blurs pin code inputs after the pin code has been changed
```

## Relevant Issues

Closes #4473

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/shared/components/inputs/Pin.svelte
+++ b/packages/shared/components/inputs/Pin.svelte
@@ -14,6 +14,7 @@
     export let glimpse = false
     export let smaller = false
     export let error
+    export let onSubmit: () => void
 
     let inputs = new Array(PIN_LENGTH)
     $: {
@@ -91,7 +92,6 @@
      */
     function changeHandlerMobile(event: Event & InputEventInit, i: number): void {
         if (!event.isTrusted || !/^[0-9]$/.test(event.data)) {
-            inputs[i] = ''
             inputElements[i].focus()
             return
         }
@@ -132,6 +132,10 @@
         if (e.target === root) {
             selectFirstEmpty()
         }
+    }
+
+    export function blur(): void {
+        inputElements.forEach((input) => input.blur())
     }
 
     export function focus(): void {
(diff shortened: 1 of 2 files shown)


packages/shared/components/inputs/Pin.svelte
```diff
@@ -14,6 +14,7 @@
     export let glimpse = false
     export let smaller = false
     export let error
+    export let onSubmit: () => void
```
amadeu2, line 17: this could be removed?

packages/shared/components/inputs/Pin.svelte
```diff
@@ -134,6 +134,10 @@
         }
     }
 
+    export function blur(): void {
```
amadeu2, line 137: this also could be removed?

amadeu2 (changes requested):
Change on `ChangePincode.svelte` fix the issue, good work! Only remove the unused code at `Pin.svelte` and we'll be ready

packages/shared/components/inputs/Pin.svelte
paul-boegelsack: I added this function to blur the sub inputs after the pin code has changed. Otherwise it should be possible to add a single digit at the end of the input you focused when submitting the new pin code. Like this:

image

packages/shared/components/inputs/Pin.svelte
paul-boegelsack: Nice catch! 👍

amadeu2 (commented):
on iOs is working as expected, but we should test it well on every android keyboard brand before merge it

PR #4504, opened 2022-09-08
enhancement: validate tag and metadata length

### LINKED ISSUE #4468: [Task]: Validate tag length on send form
### Task description

The max tag length is 64 characters, we need to validate this on the send form and any deep links.

### Requirements

- [x] Validate tag length on send form
- [x] Validate tag length on deep link

### Creation checklist

MarkNerdi:
## Summary

This PR adds validation for the tag and metadata length according to https://github.com/iotaledger/tips/blob/main/tips/TIP-0022/tip-0022.md . To be able to include inline errors, a refactoring for the `ClosableInput` was required, which added one `InputContainer` layer too much.
 

## Changelog

```
- add tag validation
- add metadata validation
- fix font size and weight on sendform for closable inputs
- refactor closable input component
```

## Relevant Issues

Closes #4468 

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/auxiliary/deep-link/handlers/wallet/operations/handleDeepLinkSendConfirmationOperation.ts
```diff
@@ -11,7 +11,7 @@ import { isValidAddressAndPrefix } from '@lib/address'
 import { openPopup } from '@lib/popup'
 
 import { SendOperationParameter } from '../../../enums'
-import { InvalidAddressError, NoAddressSpecifiedError } from '../../../errors'
+import { InvalidAddressError, MetadataLengthError, NoAddressSpecifiedError, TagLengthError } from '../../../errors'
 import { getAmountFromSearchParam } from '../../../utils'
 
 export function handleDeepLinkSendConfirmationOperation(searchParams: URLSearchParams): void {
```
MarkNerdi, line 17: i only added the validation on this handler.
IMO we should be able to reach the `SendForm`. The validation there blocks us from proceeding

packages/shared/components/inputs/ClosableInput.svelte
```diff
@@ -1,47 +1,65 @@
...
 
     let inputElement: HTMLInputElement
     let hasFocus: boolean
 
-    function onClose() {
+    $: if (inputElement === null && buttonElement) {
+        buttonElement.focus()
+    }
+
+    function handleClose() {
```
Tuditi, line 19: I'm missing a return type.

@maxwellmattryan  Should we add a coding convention to rename the button handlers like this to the following `onCloseClick()`?

```suggestion
    function handleClose(): void {
```

packages/shared/components/popups/SendFormPopup.svelte
```diff
@@ -48,6 +53,22 @@
...
+            tagError = localize('error.send.tagTooLong')
+            return Promise.reject(tagError)
+        }
+    }
+
+    let metadataError: string = ''
+    function validateMetadata(): Promise<void> {
+        if (metadata.length > 8192) {
+            metadataError = localize('error.send.metadataTooLong')
+            return Promise.reject(metadataError)
```
Tuditi, line 68: Does it make sense to throw the errors you used for deep links here instead of rejecting the promise?

packages/shared/components/popups/SendFormPopup.svelte
```diff
@@ -48,6 +53,22 @@
         isTagInputOpen = true
     }
 
+    let tagError: string = ''
+    function validateTag(): Promise<void> {
+        if (tag.length > 64) {
```
Tuditi, line 58: TIP-22 specifies the protocol size in byte length, whereas `length` returns the amount of UTF-16 (can be 2 bytes) code units. To get the bytes use something like `Buffer.from(str).size` or `new Blob([str]).size`.

A tag like `€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€€` is accepted, but is actually represented by more than 64 bytes.

Same logic applies to metadata

Ref:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length
- https://www.geeksforgeeks.org/how-to-get-the-length-of-a-string-in-bytes-in-javascript/

Tuditi (changes requested):
Nice job, I have one remark wrt validation and some questions.

packages/shared/components/popups/SendFormPopup.svelte
MarkNerdi: these errors are used for alerts and logging, but here it should only display the error under the input. i think it would require a larger change

Tuditi (changes requested):
One tiny thing overlooked, otherwise good to go!

Tuditi (approved):
LGTM! :fire:

Issue #4338, opened 2022-08-25
[Task]: Fix Stronghold error when claiming shimmer


Tuditi:
### Task description

In the claim shimmer rewards flow, stronghold sometimes fails with the following error: `Stronghold reported a procedure error: engine: fatal engine error: vault VaultId(TcAMhLekJE6GhvWh6qTYL-T2DKgaUKz7) does not exist` . It's not very consistent, so we're looking into the reproduction steps.

It seems to occur at the following line:
https://github.com/iotaledger/firefly/blob/1981da79ca7fc12e18bae6cf5b4d3cfa26a03b63/packages/shared/lib/core/account/utils/createBoundAccount.ts#L13-L14

### Requirements

TBD

### Creation checklist

- [x] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #4512, opened 2022-09-09
fix: stronghold error upon claim shimmer

Tuditi:
## Summary

This issue addresses the stronghold bug. The problem arised from a promise from the `electronApi` wasn't returned awaited properly.
...

## Changelog

```
- return Promise of copyFile
- rewrite arrow functions to regular functions
- cleanup non-existing Staking import and unused css selectors
```

## Relevant Issues

closes: #4338

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/routes/dashboard/Dashboard.svelte
```diff
@@ -24,7 +24,6 @@
     const tabs = {
         wallet: Wallet,
         settings: Settings,
-        staking: Staking,
         developer: Developer,
```
maxwellmattryan, line 27: Might be nice to have something like:
```typescript
enum AppContext {
    Developer = 'developer',
    Onboarding = 'onboarding'
    Wallet = 'wallet',
}

const tabs = {
    [AppContext.Wallet]: Wallet,
    ...
}
```

This would also work with the `en.json` locale if we add more hierarchy to the `views`, but that's definitely for another day.

Tuditi: True, but let's keep it for a separate PR, since I only removed this due to warnings that staking was missing.

maxwellmattryan:
The changes all look good. I'm guessing it's safer to return the invoke call anyway since even if it is `void` or `Promise<void>`. Might be worth switching over to `ipcRenderer.send(...)` eventually like you suggested a while back

Tuditi:
> The changes all look good. I'm guessing it's safer to return the invoke call anyway since even if it is `void` or `Promise<void>`. Might be worth switching over to `ipcRenderer.send(...)` eventually like you suggested a while back

True, but doesn't make a lot of sense to go into a lot of detail here, since we want to move away from electron.

maxwellmattryan (approved):
Works great now :fire:

PR #4526, opened 2022-09-12
fix: set system language as initial language

paul-boegelsack:
## Summary

The App still uses english as initial language. This RP changes the initial selection to the system language. 

## Changelog

```
- Selects the navigator language when the app is started for the first time
```

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -121,7 +121,8 @@
         await pollMarketData()
         await pollNetworkStatus()
 
-        const initialLocale = $appRoute === AppRoute.Welcome ? null : $appSettings.language
+        const systemLanguage = await Platform.getLanguageCode()
+        const initialLocale = $appRoute === AppRoute.Welcome ? systemLanguage : $appSettings.language
         void setupI18n({ fallbackLocale: 'en', initialLocale })
     })
 
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -122,7 +122,9 @@
         await pollNetworkStatus()
 
         const systemLanguage = await Platform.getLanguageCode()
-        const initialLocale = $appRoute === AppRoute.Welcome ? systemLanguage : $appSettings.language
+        let initialLocale = $appRoute === AppRoute.Welcome ? systemLanguage : $appSettings.language
+        // handle 'es' & 'pt' cases when we can't get the region code
+        initialLocale = initialLocale === 'es' ? 'es-ES' : initialLocale === 'pt' ? 'pt-PT' : initialLocale
         void setupI18n({ fallbackLocale: 'en', initialLocale })
     })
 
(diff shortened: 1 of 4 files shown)


packages/mobile/App.svelte
```diff
@@ -122,6 +120,9 @@
         await getVersionDetails()
         await pollMarketData()
         await pollNetworkStatus()
+
+        const initialLocale = $appRoute === AppRoute.Welcome ? null : $appSettings.language
```
amadeu2, line 124: I think still we need a way to detect the device language, maybe we could use [device.getlanguagecode()](https://capacitorjs.com/docs/apis/device#getlanguagecode)
With the capacitor v4 we also could use `device#getlanguageTag` to process also the region code and possibly apply to the amount input to handle the `,.` issue

amadeu2 (commented):
When tested with a clean install I still getting English as default, having Spanish defined on my device, maybe using the capacitor device plugin could help us to fix it

packages/mobile/App.svelte
paul-boegelsack: > When tested with a clean install I still getting English as default, having Spanish defined on my device, maybe using the capacitor device plugin could help us to fix it

Yeah, seems there is still an issue. It works on my iOS device but not on may Android it seems, just checked again. Do you have the problem on both OS?

amadeu2: To reproduce on iOS we need to add a new language and keep English or any other there should be more than 1, then uninstall the app, change the language and install again. The JS way to get the language returns the first language of the list, but not the selected one on the OS. With Capacitor device plugin we get the correct selected language

amadeu2 (changes requested):
I think adding only the [device.getlanguagecode()](https://capacitorjs.com/docs/apis/device#getlanguagecode) would be enough, forget by now the Capacitor v4 update, I've been testing on this branch some time ago but we'll need to check some issues https://github.com/iotaledger/firefly/tree/chore/update-capacitor-v4

packages/mobile/App.svelte
paul-boegelsack: > I think adding only the [device.getlanguagecode()](https://capacitorjs.com/docs/apis/device#getlanguagecode) would be enough, forget by now the Capacitor v4 update, I've been testing on this branch some time ago but we'll need to check some issues https://github.com/iotaledger/firefly/tree/chore/update-capacitor-v4

Alright, I added the getLanguageCode function and it seems to work know 👍 Could you check again? 😄

amadeu2: now works! but seems there is still some issue with 'es' and 'pt' lang codes, I'll fix it

amadeu2 (approved):
Tested and working nice on Android and iOS 🥳 fixed specific cases on 'es' and 'pt' region codes

PR #4558, opened 2022-09-15
fix: splash screen delays

### LINKED ISSUE #4532: [Task]: Fix splash screen delays
### Task description

On both Android and iOS the splash screen behaviour could be improved. On app open, the splash screen should display until the login screen is ready to render. 

### Requirements

There should be no gap between the splash screen closing and login screen rendering.

### Creation checklist

paul-boegelsack:
## Summary

This PR removes the empty view that appears for some seconds when the splash screen hides.

## Changelog

```
- Changes the conditions that checks if the splash screen should be hidden
```

## Relevant Issues

Closes #4532

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [x] iOS
  - [x] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/mobile/App.svelte
+++ b/packages/mobile/App.svelte
@@ -72,7 +72,7 @@
         document.dir = $localeDirection
     }
 
-    $: if ($isLocaleLoaded) {
+    $: if ($isLocaleLoaded && showSplash) {
         void hideSplashScreen()
     }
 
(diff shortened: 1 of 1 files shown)


packages/mobile/App.svelte
```diff
@@ -128,78 +128,76 @@
...
-        <Route route={AppRoute.Login}>
-            <Login locale={$_} {goto} />
-        </Route>
-        <ToastContainer />
-    </div>
-    <div class="scanner-ui">
-        <QRScanner locale={$_} />
-    </div>
-{/if}
+<!-- empty div to avoid auto-purge removing dark classes -->
```
begonaalvarezd, line 131: we might need to add here or wrapping the relevant routes a `#key $isLocaleLoaded` because with these removals the localization is not propagating properly when you change language, things dont translate properly until you close the app and start again

begonaalvarezd (dismissed):
There is an improvement indeed, tested on android and works great. Added a coment though because changing the language is not updating correctly 🙏🏼

paul-boegelsack:
> There is an improvement indeed, tested on android and works great. Added a coment though because changing the language is not updating correctly 🙏🏼

Thanks for catching this, I re added the isLocaleLoaded condition 🙏

amadeu2:
To test it well we need to build a production release and open / close the app from the device to mesure the init times.
On Android, we still need to wait the 3 seconds only for onboarding routes to avoid language jumps, we should remove de timeout only for the login route.
On iOS on my iPhone 7 there are no performance issues even at onboarding, so we could apply the delay only for Android.

amadeu2 (changes requested):
To test it well we need to build a production release and open / close the app from the device to mesure the init times.
On Android, we still need to wait the 3 seconds only for onboarding routes to avoid language jumps even on high end devices as Pixel 3, we should remove de timeout only for the login route.
On iOS on my iPhone 7 there are no performance issues even at onboarding, so we could apply the delay only for Android.

begonaalvarezd:
> To test it well we need to build a production release and open / close the app from the device to mesure the init times. On Android, we still need to wait the 3 seconds only for onboarding routes to avoid language jumps even on high end devices as Pixel 3, we should remove de timeout only for the login route. On iOS on my iPhone 7 there are no performance issues even at onboarding, so we could apply the delay only for Android.

I tested installing an alpha on my phone, closing and opening the app i installed, woudlnt that be enough to test?

paul-boegelsack:
> To test it well we need to build a production release and open / close the app from the device to mesure the init times. On Android, we still need to wait the 3 seconds only for onboarding routes to avoid language jumps even on high end devices as Pixel 3, we should remove de timeout only for the login route. On iOS on my iPhone 7 there are no performance issues even at onboarding, so we could apply the delay only for Android.

We should definitely take this into account when setting up mobile for Shimmer. For now I would just fixing the blank view after the splash screen. Does the blank screen still appear on your devices?

amadeu2 (approved):
I've made some fixes, tested and working as expected on Android and iOS.
We only needed the delay for Android devices, since iOS has not jump or glitch issues at loading locales. Also we need it only at Welcome screen, as at Login is not needed.

PR #4611, opened 2022-09-20
refactor: cleanup crash reporting

nicole-obrien:
## Summary

...

## Changelog

```
- Remove unused store
- Move default app settings
- Set default crash reporting to true
- Set default deeplinking to true
```

## Relevant Issues

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/app/constants/default-app-settings.constant.ts
```diff
@@ -0,0 +1,11 @@
+import { AppTheme } from '../enums'
+import { IAppSettings } from '../interfaces'
+
+export const DEFAULT_APP_SETTINGS: IAppSettings = {
+    deepLinking: true,
```
kraftjs, line 5: Shouldn't we keep deepLinking `false` and request deepLink permission first?

nicole-obrien: I actually don't even know if it's used any more. The setting doesn't make sense because it doesn't do anything with registering a url schema with the OS. So any permissions about what protocol is registered with the OS is managed by the OS. We should in future look at allowing the user to register and deregister through the app. But false doesn't do anything (at least previously) because the url schema is still registered and instead of following the link it just takes you to settings?

kraftjs: Hmmm, I guess I'm not too familiar with how the deep links work under the hood because the part about url schemas and the OS is going over my head

...but this changed DEFAULT_APP_SETTINGS does change the default value of the deepLink setting in newly created profiles (just checked)

packages/shared/lib/core/app/constants/default-app-settings.constant.ts
```diff
@@ -0,0 +1,11 @@
...
+    language: 'en',
+    theme: AppTheme.Light,
+    darkMode: false,
+    notifications: true,
+    sendCrashReports: true,
```
kraftjs, line 10: Just checked this via `yarn start`. Newly created profiles still have a false value for the "Crash reports" setting (Send crash reports is untoggled)

packages/shared/lib/core/app/constants/default-app-settings.constant.ts
nicole-obrien: Let me revert that change and we can make it in a separate Pr to be clear what we are doing with that setting

nicole-obrien: As for new profiles this change will only affect fresh installs, or you can delete the settings object from application data to test

packages/shared/lib/core/app/constants/default-app-settings.constant.ts
nicole-obrien: Thats because this is app settings so only works at fresh install, or if you delete at minimum the settings from the application data

kraftjs (approved):
LGTM, can confirm crashReporting is reset after deleting the Electron folder and creating a new profile

PR #4667, opened 2022-09-23
enhancement: hide pointless activities

### LINKED ISSUE #4648: [Task]: Hide pointless activity where amount = return amount and rest of output is empty
### Task description

People can spam us with basic outputs where the return amount = the amount and there is no native tokens, there is no point claiming it as all the funds are just being sent back to the sender.

### Requirements

- [ ] Do not build activity for this specific case where there is no additional assets in the ouput and the amount = the return amount

### Creation checklist

MarkNerdi:
## Summary
If a transactions amount is the same as the storage deposit return, it will give 0 funds to the user if they claim it.  

## Changelog

```
- dont show activities which contain 0 funds
```

## Relevant Issues
Closes #4648

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


nicole-obrien (changes requested):
You haven't done this check when we receive a new output while logged in?

nicole-obrien (changes requested):
Thinking about this, I think this should be done as a filter that the user can toggle. What do you think?

MarkNerdi:
we could set them to `hidden` and make them visible if we select the `show hidden` filter

nicole-obrien (changes requested):
So currently this is applying to all outputs if they are async, we only want to target the basic outputs that dont have native tokens, as other outputs are inherently an object/asset.

Also we should separate the filter so there is a show valueless activities filter. And a show all hidden filter.

Tuditi (approved):
LGTM! :fire:

PR #4775, opened 2022-09-29
fix: clear the onboarding and active profile before onboarding starts

### LINKED ISSUE #4765: [Task]: Clear the onboarding and active profile before onboarding starts
### Task description

In some cases the active profile or onboarding profile could be present before onboarding flow starts, and this could lead to problems.

### Requirements

- [x] Clear onboarding profile when...
    - [x] Entering onboarding flow
    - [x] Completing onboarding flow 
- [x] Clear active profile when...
    - [x] Logging out
    - [x] Entering onboarding flow 
- [x] (Opt.) Refactor a function for initiating the onboarding flow to be used in the multiple places the flow starts

### Creation checklist

maxwellmattryan:
## Summary

...

## Changelog

```
```

## Relevant Issues

Closes #4765 

### Platforms

- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/desktop/App.svelte
+++ b/packages/desktop/App.svelte
@@ -120,15 +120,15 @@
         })
         Electron.onEvent('menu-create-developer-profile', () => {
             get(appRouter).reset()
-            initialiseOnboardingProfile(true, NetworkProtocol.Shimmer)
             initialiseOnboardingRouters()
+            initialiseOnboardingProfile(true, NetworkProtocol.Shimmer, true)
             get(appRouter).next({ shouldAddProfile: true })
         })
         Electron.onEvent('menu-create-normal-profile', () => {
             get(appRouter).reset()
-            initialiseOnboardingProfile(false, NetworkProtocol.Shimmer)
-            updateOnboardingProfile({ networkType: NetworkType.Mainnet })
             initialiseOnboardingRouters()
+            initialiseOnboardingProfile(false, NetworkProtocol.Shimmer, true)
+            updateOnboardingProfile({ networkType: NetworkType.Mainnet })
             get(appRouter).next({ shouldAddProfile: true })
         })
         Electron.hookErrorLogger((err) => {
--- a/packages/shared/lib/contexts/onboarding/actions/initialiseOnboardingProfile.ts
+++ b/packages/shared/lib/contexts/onboarding/actions/initialiseOnboardingProfile.ts
@@ -1,16 +1,30 @@
-import { NetworkProtocol } from '@core/network'
-import { profileManager } from '@core/profile-manager'
 import { get } from 'svelte/store'
+
+import { stopPollingLedgerNanoStatus } from '@core/ledger'
+import { NetworkProtocol } from '@core/network'
+import { destroyProfileManager, profileManager } from '@core/profile-manager'
+
 import { OnboardingProfileManagerAlreadyInitializedError } from '../errors'
 import { buildOnboardingProfile } from '../helpers'
-import { onboardingProfile } from '../stores'
+import { isOnboardingLedgerProfile, onboardingProfile } from '../stores'
 
 /**
  * Builds a new onboarding profile and sets the Svelte store accordingly.
  */
-export function initialiseOnboardingProfile(isDeveloperProfile: boolean, networkProtocol?: NetworkProtocol): void {
+export function initialiseOnboardingProfile(
+    isDeveloperProfile: boolean,
+    networkProtocol?: NetworkProtocol,
+    overrideExistingProfileManager = false
+): void {
     if (get(profileManager)) {
-        throw new OnboardingProfileManagerAlreadyInitializedError()
+        if (overrideExistingProfileManager) {
+            if (get(isOnboardingLedgerProfile)) {
+                stopPollingLedgerNanoStatus()
+            }
+            destroyProfileManager()
+        } else {
+            throw new OnboardingProfileManagerAlreadyInitializedError()
+        }
     }
 
     const _newProfile = buildOnboardingProfile(isDeveloperProfile, networkProtocol)
--- a/packages/shared/lib/core/profile/actions/active-profile/login.ts
+++ b/packages/shared/lib/core/profile/actions/active-profile/login.ts
@@ -119,7 +119,7 @@ export async function login(loginOptions?: ILoginOptions): Promise<void> {
     } catch (err) {
         handleError(err)
         if (!loginOptions?.isFromOnboardingFlow) {
-            await logout()
+            await logout(true)
         }
         _loginRouter.previous()
         resetLoginProgress()
--- a/packages/shared/components/Idle.svelte
+++ b/packages/shared/components/Idle.svelte
@@ -54,7 +54,7 @@
     }
 
     function lock(): void {
-        void logout()
+        void logout(true)
     }
 
     onDestroy(() => {
--- a/packages/shared/routes/dashboard/Dashboard.svelte
+++ b/packages/shared/routes/dashboard/Dashboard.svelte
@@ -32,7 +32,7 @@
 
     onMount(() => {
         Platform.onEvent('menu-logout', () => {
-            void logout()
+            void logout(true)
         })
 
         Platform.onEvent('deep-link-params', (data: string) => {


packages/shared/lib/core/profile/actions/active-profile/login.ts
```diff
@@ -119,7 +119,7 @@ export async function login(loginOptions?: ILoginOptions): Promise<void> {
     } catch (err) {
         handleError(err)
         if (!loginOptions?.isFromOnboardingFlow) {
-            await logout()
+            await logout(true)
```
MarkNerdi, line 122: it seems that we always want to `clearActiveProfile`. Maybe we should set its default to `true` or fully remove the param

packages/shared/lib/contexts/onboarding/actions/initialiseOnboardingProfile.ts
```diff
@@ -1,16 +1,30 @@
...
+import { isOnboardingLedgerProfile, onboardingProfile } from '../stores'
 
 /**
  * Builds a new onboarding profile and sets the Svelte store accordingly.
  */
-export function initialiseOnboardingProfile(isDeveloperProfile: boolean, networkProtocol?: NetworkProtocol): void {
+export function initialiseOnboardingProfile(
+    isDeveloperProfile: boolean,
+    networkProtocol?: NetworkProtocol,
+    overrideExistingProfileManager = false
```
MarkNerdi, line 17: is there a case where we wouldn't want to override the old manager?

MarkNerdi (changes requested):
Works nice!
left some comments about (probably) redundant parameters

packages/shared/lib/core/profile/actions/active-profile/login.ts
nicole-obrien: Ahh I see that it was added to idle will comment because we don't want it cleared on auto logout. Perhaps another place to make sure it is cleared is when we press back on the pin input screen.

nicole-obrien: We don't actually want to clear the active profile here either? Because it has already been created so if there is an error logging in we want them to go back to the login screen from where they came and this requires the active profile.

nicole-obrien (changes requested):
I think this PR has missed the requirements of the task at hand. We do not want to clear the active profile in certain cases such as logging out on error of login, or auto logging out. As the active profile has a valid need on the Enter pin screen of login.

Instead of trying to make sure that we are "cleaning down" the active profile, we need to make sure everything is reset when initating the onboarding flow from the various places it can be initiated.

packages/shared/lib/contexts/onboarding/actions/initialiseOnboardingProfile.ts
maxwellmattryan: There are some spots in the onboarding where this is called without overriding the existing manager. It would be worth going through onboarding and cleaning this logic up a bit more from the first refactoring at some point.

packages/desktop/App.svelte
```diff
@@ -112,15 +112,21 @@
...
         Electron.onEvent('menu-create-developer-profile', () => {
             get(appRouter).reset()
-            initialiseOnboardingProfile(true, NetworkProtocol.Shimmer)
-            initialiseOnboardingRouters()
+            void initialiseOnboarding({
+                isDeveloperProfile: true,
+                networkProtocol: NetworkProtocol.Shimmer,
+                resetRouters: true,
+            })
             get(appRouter).next({ shouldAddProfile: true })
```
nicole-obrien, line 120: We should explicitly state what route to go to in the initialiseOnboarding function

packages/shared/lib/core/profile/actions/active-profile/logout.ts
```diff
@@ -15,18 +15,19 @@ import {
...
-export function logout(clearActiveProfile: boolean = false, _lockStronghold: boolean = true): Promise<void> {
+export function logout(clearActiveProfile: boolean = true, _lockStronghold: boolean = true): Promise<void> {
     const { lastActiveAt, loggedIn, hasLoadedAccounts, type } = get(activeProfile)
 
     // (TODO): Figure out why we are using a promise here?
     return new Promise((resolve) => {
         if (get(isSoftwareProfile)) {
             _lockStronghold && lockStronghold()
-        } else if (isLedgerProfile(type)) {
+        } else if (isLedgerProfile(type) || get(isOnboardingLedgerProfile)) {
```
nicole-obrien, line 30: This function is to logout of the active profile and is in the profile/actions/active-profile context. So if we need to clear down anything for onboarding then this should be outside of the logout function.

packages/shared/lib/contexts/onboarding/actions/index.ts
```diff
@@ -14,6 +14,7 @@ export * from './syncShimmerClaimingAccount'
 export * from './generateMnemonicForOnboardingProfile'
 export * from './handleTransactionInclusionEventForShimmerClaiming'
 export * from './initialiseFirstShimmerClaimingAccount'
+export * from './initialiseOnboarding'
```
nicole-obrien, line 17: Perhaps this file should be initialiseOnboardingFlow as it is going to navigate the user to the start of the onboarding flow.

packages/shared/lib/contexts/onboarding/actions/initialiseOnboarding.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    await logout()
+
+    initialiseOnboardingProfile(isDeveloperProfile, networkProtocol, true)
+
+    if (networkType) {
+        updateOnboardingProfile({ networkType })
+    }
+
+    if (resetRouters) {
```
nicole-obrien, line 22: When would we not reset the router? Can we not just do this in all cases as we want to be at the start of the correct onboarding flow no matter what.

packages/shared/lib/contexts/onboarding/actions/initialiseOnboardingProfile.ts
```diff
@@ -1,16 +1,30 @@
...
+): void {
     if (get(profileManager)) {
-        throw new OnboardingProfileManagerAlreadyInitializedError()
+        if (overrideExistingProfileManager) {
```
nicole-obrien, line 20: Should this clearing down not be done in a clear onboarding function called in intialiseOnboarding

nicole-obrien (changes requested):
Nearly there  still using a couple of functions to start the onboarding flow instead of using the main initialiseOnboarding function you have created. I have detailed all the requirements here to make it easier to fulfiul.

Requirements
1 function to initialise onboarding flow in all locations
 Locations of onboarding entry
* User enters app and goes from app setup -> onboarding
* User enters app having completed app setup and starts in onboarding
* User creates new profile from profile selection view
* User creates developer profile from electron menu
* User creates normal profile from electron menu

Two starting points within onboarding which can be automatic or manually set
* Profile setup view if is developer is false
    * Normal profile through electron menu
    * Or Default flows on production app
    * Network protocol = Shimmer, Network type = Mainnet
* Network type view if is developer is true
    * Developer profile through electron menu
    * Or Default flows on develop, alpha and beta app
    * Network protocol = Shimmer, Network type = undefined

nicole-obrien (commented):
I don't see any changes to intitilisation during the first onboarding, i.e. after app setup is complete

packages/desktop/electron/preload.js
```diff
@@ -55,7 +55,7 @@ try {
             levelFilter: 'debug',
             targetExclusions: ['h2', 'hyper', 'rustls', 'message_handler'],
         }
-        WalletApi.initLogger(loggerOptions)
+        // WalletApi.initLogger(loggerOptions)
```
MarkNerdi, line 58: this will remove the console error, but we won't have any logging then, right?
I think its an ok trade-off, because if we need wallet.rs logs, we can then just enable it

MarkNerdi (changes requested):
One change needs to be made do be able to compile it

packages/desktop/electron/preload.js
Tuditi: If there is no explicit reason for them to be commented out, it's best to have them in.

packages/desktop/electron/preload.js
```diff
@@ -55,7 +55,7 @@ try {
             levelFilter: 'debug',
             targetExclusions: ['h2', 'hyper', 'rustls', 'message_handler'],
         }
-        WalletApi.initLogger(loggerOptions)
```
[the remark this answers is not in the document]
maxwellmattryan: I think I forgot to uncomment it back in

packages/desktop/electron/preload.js
maxwellmattryan: I forgot to uncomment it back in :sweat_smile:

MarkNerdi (approved):
LGTM!

PR #4779, opened 2022-09-30
fix: NumberInput

### LINKED ISSUE #4769: [Task] AmountInput allows ` ^ `
The AssetAmountInput allows the ^ - sign as an input, which then breaks the whole app because `big.js` cannot convert that to a number

jeeanribeiro:
## Summary
Number input needs to have the native HTML `type=number`, otherwise it leads to problems in parsing numbers as it is happening with big.js
So in this PR I've changed the component to not be a wrapper for TextInput, but have the same customizability as it

## Changelog
```markdown
- Changes from `TextInput` to `Input` in `NumberInput`
- Adds text props in `NumberInput`
- Improves code readability in `TextInput`
```

## Relevant Issues
closes #4769 

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions
- Try to type `.` in AssetAmountInput
- If the app is still interactable so the fix has worked

## Checklist

### DIFF
--- a/packages/shared/components/inputs/NumberInput.svelte
+++ b/packages/shared/components/inputs/NumberInput.svelte
@@ -1,24 +1,39 @@
 <script lang="typescript">
-    import { TextInput } from 'shared/components'
     import { localize } from '@core/i18n'
+    import Input from './Input.svelte'
+    import { FontWeight, TextPropTypes, TextType } from 'shared/components/Text.svelte'
 
     export let inputElement: HTMLInputElement
-
-    export let disabled = false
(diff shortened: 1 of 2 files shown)


MarkNerdi (approved):
the ^ is displayed on first keypress, but disappears when it would be added to the numberstring because of the property of ^ ` and chars like this.
I think we cannot  avoid that, so GOOD TO GO

Issue #4777, opened 2022-09-30
[Task]: Upgrade wallet.rs to `2.0.3-rc.1`


nicole-obrien:
### Task description

Latest wallet.rs with getAddressIndexs and making meta private has been released.

### Requirements

- [ ] Update wallet.rs to 2.0.3-rc.1
- [ ] Remove uses of account.meta
- [ ] Use getAccountIndexes instead of getAccounts
- [ ] Resolve other change logs

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #4807, opened 2022-10-05
chore: upgrade wallet rs

Tuditi:
## Summary

Upgraded wallet.rs to contain latest changes in line with 2.0.3-rc1

...

## Changelog

```
- Updated wallet.rs to 2.0.3-rc.1
- Added extra methods to contextBridge
- Replace meta with getMetadata
- Remove list prefixes from wallet.rs methods
- Updated error types to be camelCase
- Removed duplicate error declarations
- Replaced deleteAccountsAndDatabase with removeProfileFolder
- Removed unused api calls
- Removed duplicate createAccount (name was createBoundAccount)
```

## Relevant Issues

closes: #4777 

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


nicole-obrien (dismissed):
As mentioned on the call, I think we should be reducing the calls to the the metadata / alias etc because we can load that once when we load the account and then just use the account.name or account.id from local state.

Additionally with the change to bridging the methods properly, I wonder if you can recursively do this so that we don't have to separate getAccounts and recoverAccounts in the preload file.

Also with the change to addresses we can no longer rely on the first address being passed to us using the addresses api is the first address index that is external. We may need to search this response for the correct address and then save this in the account state so we only do this once during account loading. However, it might make more sense to have a specific endpoint for getting a specific addresses and then we can pass in the first address index for this.

packages/shared/lib/core/profile-manager/api/createAccount.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { IAccount } from '@core/account'
+
+import { api, IProfileManager, profileManager } from '..'
+
+export function createAccount(
+    payload: CreateAccountPayload,
+    manager: Writable<IProfileManager> = profileManager
+): Promise<IAccount> {
+    const { id } = get(manager)
```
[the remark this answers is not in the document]
Tuditi: Because we use this function, while we don't use the others :raccoon: :D

Tuditi:
> As mentioned on the call, I think we should be reducing the calls to the the metadata / alias etc because we can load that once when we load the account and then just use the account.name or account.id from local state.
> 
> Additionally with the change to bridging the methods properly, I wonder if you can recursively do this so that we don't have to separate getAccounts and recoverAccounts in the preload file.
> 
> Also with the change to addresses we can no longer rely on the first address being passed to us using the addresses api is the first address index that is external. We may need to search this response for the correct address and then save this in the account state so we only do this once during account loading. However, it might make more sense to have a specific endpoint for getting a specific addresses and then we can pass in the first address index for this.

I addressed all your issues, except the one about binding the object methods recursively. I experimented with it and I don't think it's possible. The main difficulty is that we don't have a generic way of retrieving the objects returned by specific methods.

packages/shared/lib/contexts/onboarding/actions/claimShimmerRewards.ts
```diff
@@ -60,7 +60,7 @@ async function claimShimmerRewardsForShimmerClaimingAccounts(
 async function claimShimmerRewardsForShimmerClaimingAccount(
     shimmerClaimingAccount: IShimmerClaimingAccount
 ): Promise<void> {
-    const recipientAddress = shimmerClaimingAccount?.twinAccount?.meta?.publicAddresses[0]?.address
+    const recipientAddress = (await shimmerClaimingAccount?.twinAccount?.addresses())[0].address
```
MarkNerdi, line 63: do we need the following?
(await shimmerClaimingAccount?.twinAccount?.addresses())?.[0].address

Tuditi: Good catch, we need the first address which is external. I added a utils function to account that fetches this, since we use it in two places.

MarkNerdi (approved):

PR #4834, opened 2022-10-10
feat: add alias selection on minting

### LINKED ISSUE #4117: [Task]: Add confirmation popup and error handling for minting native tokens
### Task description

We should copy the flow of sending considering we are technically sending a when minting we should add  a confirmation page with error handling.

### Requirements

- [x] Confirmation page for minting flow
- [x] Error handling

### Creation checklist

MarkNerdi:
## Summary
This PR introduces a Alias selection on the native token minting form as well as a confirmation page.
If the user has no alias, they get prompted to create one 

## Changelog

```
- Add alias selector + validation to MintNativeTokenForm
- add confirmation page for token minting
- add warning popup if no alias exists
```

## Relevant Issues

Closes #4117 

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/components/modals/AliasSelector.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+    export let modal: Modal = undefined
+    export let selected: string = undefined
+    export let onClose: () => void
+
+    $: aliasIds = get(selectedAccountActivities)
+        .filter((activity) => activity.type === ActivityType.Alias)
+        .map((activity) => {
+            const aliasId = (activity.data as IAliasActivityData).aliasId
+            return { value: aliasId, label: truncateString(aliasId, 9, 9) }
```
Tuditi, line 16: Why do you get the `aliasId` from the `selectedAccountActivities`? It feels like this could error, what if you transferred control over an alias output already. It would be part of your activities, but you would still need to create one. I think it's better to get this info from wallet.rs.

packages/shared/components/popups/MintNativeTokenConfirmationPopup.svelte
```diff
@@ -0,0 +1,120 @@
...
+                ...(url && { url }),
+                ...(logoUrl && { logoUrl }),
+            })
+            closePopup()
+        } catch (reason) {
+            handleError(reason.error)
+        }
+    }
+
+    function handleBack() {
```
Tuditi, line 69: missing return type

packages/shared/components/popups/MintNativeTokenConfirmationPopup.svelte
```diff
@@ -0,0 +1,120 @@
...
+
+    async function handleMint(): Promise<void> {
+        try {
+            await checkActiveProfileAuth(mintAction, { stronghold: true, ledger: false })
+        } catch (reason) {
+            handleError(reason.error)
+        }
+    }
+
+    onMount(async () => {
```
Tuditi, line 85: I don't think onMount is allowed to be async

packages/shared/locales/en.json
```diff
@@ -887,21 +887,27 @@
...
                 "url": "URL",
                 "logoUrl": "Logo URL",
-                "logo": "Logo"
+                "logo": "Logo",
+                "alias": "Alias"
             }
         },
+        "noAlias": {
+            "title": "Unable to mint",
+            "description": "You must own an alias output to mint native tokens. To continue, please first create an alias"
```
Tuditi, line 909:
```suggestion
            "description": "You must own an alias output to mint native tokens. To continue, please create an alias first"
```

packages/shared/components/atoms/buttons/MintNativeTokenButton.svelte
```diff
@@ -1,12 +1,34 @@
...
+    import { ActivityType, selectedAccountActivities } from '@core/wallet'
+    import { closePopup, openPopup } from '@lib/popup'
     import { OnboardingButton } from 'shared/components'
+    import { get } from 'svelte/store'
 
-    function handleMintNativeToken() {
-        openPopup({
-            type: 'mintNativeTokenForm',
-        })
+    $: aliases = get(selectedAccountActivities).filter((activity) => activity.type === ActivityType.Alias)
```
Tuditi, line 8: This should be gotten from wallet.rs (`account.getBalance()`), since you could lose control of an alias in the future

Tuditi (changes requested):
Code looks very good, but I have some minor questions. 

Also I have an undefined error if I try to create an alias without having SMR. This should be handled appropriately.

packages/shared/components/popups/MintNativeTokenConfirmationPopup.svelte
MarkNerdi: it is, but its not being awaited

packages/shared/components/modals/AliasSelector.svelte
MarkNerdi: good point!

Tuditi (approved):
LGTM! :fire:

Issue #4926, opened 2022-10-19
[Task]: Create Network Selector Dropdown


Tuditi:
### Task description

TBD

### Requirements

TBD

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #4982, opened 2022-10-19
feat: network selector on send

Tuditi:
## Summary

Added a network selector to the send form popup

...

## Changelog

```
- Added a network selector
```

## Relevant Issues

closes: #4926 

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


packages/shared/components/molecules/SelectorInput.svelte
```diff
@@ -2,7 +2,7 @@
     import { localize } from '@core/i18n'
     import { InputContainer, Modal, TextInput } from 'shared/components'
 
-    export let value: string
+    export let value: string = undefined
```
maxwellmattryan, line 5: Does it make more sense to use `undefined` here rather than an empty string `''`? Both are considered falsy

packages/shared/lib/core/network/constants/destination-network-address.constant.ts
```diff
@@ -0,0 +1,6 @@
+import { DestinationNetwork } from '../enums'
+
+export const DESTINATION_NETWORK_ADDRESS: Readonly<{ [key in DestinationNetwork]?: string }> = {
+    [DestinationNetwork.Shimmer]: '-',
+    [DestinationNetwork.ShimmerEvm]: 'rms1qrut5ajyfrtgjs325kd9chwfwyyy2z3fewy4vgy0vvdtf2pr8prg5u3zwjn',
```
maxwellmattryan, line 5: Probably would be better to have a `const` object with both destination network addresses for testnet EVM _and_ ShimmerEVM.

packages/shared/lib/core/network/enums/destination-network.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum DestinationNetwork {
+    Shimmer = 'Shimmer',
+    ShimmerEvm = 'Shimmer EVM',
```
maxwellmattryan, line 3: I suppose this doesn't need to adhere to the normal `enum` conventions since the text doesn't need to be localized?

maxwellmattryan (changes requested):
Left some specific comments / questions, but generally:

- Not sure if related to this PR but the error isn't displayed in the UI (beneath the address input) whenever no address is entered and you try to click "Send"
- Definitely not related to this PR, but the Figma designs say "Recipient" whereas currently it's saying "Alias" for the address inputs below the network selector.

Code and UI looks good though!

packages/shared/components/inputs/NetworkInput.svelte
```diff
@@ -0,0 +1,21 @@
...
+    let modal: Modal = undefined
+</script>
+
+<SelectorInput
+    labelLocale="general.destinationNetwork"
+    bind:value={network}
+    bind:inputElement
+    bind:modal
+    {error}
+    readonly
```
MarkNerdi, line 18: could you update the pointer event if its `readonly`

packages/shared/components/modals/NetworkSelector.svelte
```diff
@@ -0,0 +1,36 @@
...
+                >
+                    <Text type={TextType.pre} fontSize="sm" color="gray-800">{network}</Text>
+                    <Text type={TextType.pre} fontSize="sm" color="gray-600">
+                        {truncateString(NETWORK_ADDRESS[$activeProfile.networkType][network], 6, 6)}
+                    </Text>
+                </button>
+            {/each}
+        </network-picker-modal>
+    </Modal>
+{/if}
```
MarkNerdi, line 36: i think we could put the follwing into the selector component and only make the item content to a slot
```
#if list.length > 0}
    <Modal bind:this={modal} position={{ left: '0', top: '100%' }} classes="w-full p-4" on:close={onClose}>
        <network-picker-modal class="max-h-64 flex flex-col space-y-1 scrollable-y" in:fade={{ duration: 100 }}>
            {#each list as item}
                <button
                    on:click={() => onClick(item)}
                    class="w-full flex flex-row flex-1 justify-between px-2 py-3 rounded-lg hover:bg-blue-50 dark:hover:bg-gray-800 dark:hover:bg-opacity-20"
                >
                    <slot />
                </button>
            {/each}
        </network-picker-modal>
    </Modal>
{/if}
```

MarkNerdi (commented):
Works and looks really good, 
left a few small comments

packages/shared/components/inputs/NetworkInput.svelte
Tuditi: I added it to the selector input. It's not perfect, but a proper fix would require extracting the button type from the `InputContainer` and that should be part of a wider refactor.

MarkNerdi (approved):

PR #5002, opened 2022-10-21
feat:add sending l2 tx

Tuditi:
## Summary

...

## Changelog

```
```

## Relevant Issues

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


nicole-obrien (changes requested):
As discussed on the call heres the main improvements to make:

- Lets make it so that we are not converting anything until we actually call the send function, so we can store all the params in the new transaction store and go backwards and forwards with ease.
- Add a network subject type to store network and address or just address as the subject and then store this in the recipient of the new transaction
- Add a layer 2 params to the new transaction and potentially a flag to say its layer 2. Then in another task we can easily use this data to go backwards and forwards in the send flow or to show the layer 2 details rather than the metadata as hex.
- Add a feature flag to make the network field readonly in the mainnet profiles but not in dev profiles.
- Add a Shimmer Testnet EVM network so its different to the Shimmer EVM

nicole-obrien (changes requested):
As per our review call:

- some of file names for constants don't match the constant names
- we should move the smart contract params to the params of the function to make it fully generic
- we should rename getLayer2Metadata to make it clear this is for a specific contract and call
- Ideally we should be adding the gas budget onto the amount when creating the output as this is an addition to the user defined amount which is used in the allowance field of the contract

nicole-obrien (approved):
LGTM!

PR #5018, opened 2022-10-25
enhancement: merge send forms

### LINKED ISSUE #4918: [Task]: Add Gifting Storage Deposit on NFT Send

### LINKED ISSUE #4978: [Task]: Merge Send NFT Popups with the normal Send Flow Popups
We want  to update the sendflow, such that the initial `SendFormPopup` contains tabs to switch between `TokenTransfer` and `NftTransfer`. 
For this we need to extract both Forms of the current Popups and then use them in the combined SendForm. 

For the SendConfirmationPopup, we can use the same approach we have in the `ActivityDetailsPopup` which handles different activity types

MarkNerdi:
## Summary
This PR combines the sendflows for tokens and for nfts into one popup. The Form popup includes tabs to switch between the 2 options, the Confirmation popup displays things differently corresponding to the previously selected option.

## Changelog

```
- Reset Sendbutton
- extract Token and Nft send form into own components
- unify newTransactionDetails store
- unify SendFormPopup
- unify SendConfirmationPopup
```

## Relevant Issues

Closes #4918 
Closes #4978

### Platforms

- __Desktop__
  - [ ] MacOS
  - [ ] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

...

## Checklist


Tuditi (commented):
Looks good overall, left some minor comments :relaxed:

packages/shared/lib/core/wallet/interfaces/nft-activity-data.interface.ts
```diff
@@ -10,9 +10,11 @@ export interface INftActivityData {
     type: ActivityType.Nft
     outputId: string
     nftId: string
-    metadata: string
+    nftMetadata: string
```
Tuditi, line 13: Why is nftMetadata different from metadata?

packages/shared/lib/core/wallet/types/new-transaction-details.type.ts
```diff
@@ -0,0 +1,33 @@
...
+    asset: IAsset
+    unit: string
+    metadata: string
+    recipient: Subject
+    tag: string
+    expirationDate?: Date
+    giftStorageDeposit?: boolean
+    surplus?: string
+    disableToggleGift?: boolean
+    disableChangeExpiration?: boolean
```
Tuditi, line 18: Maybe have a common TransactionDetails type that is extended in this file to avoid duplication?

packages/shared/lib/core/wallet/utils/buildNftOutputData.ts
```diff
@@ -1,15 +1,26 @@
...
     expirationDate: Date,
     nftId: string,
     immutableFeatures: FeatureTypes[],
+    metadata: string,
+    tag: string,
     recipientAddress: string,
-    accountAddress: string
+    accountAddress: string,
+    storageDeposit: string,
+    giftStorageDeposit: boolean
```
Tuditi, line 23: I don't like this, the multitude of arguments should be converted into a couple of meaningful objects. Check Clean Code p.40 (https://github.com/dev-marko/clean-code-book/blob/master/Clean%20Code%20(%20PDFDrive.com%20).pdf)

image

Tuditi (changes requested):
Looks good, I left some comments. There is also an annoying glitch present if you try to add metadata/tag to a transaction. For a split second the optional buttons are above each other instead of under each other. Not sure if it's your PR that introduced it. If it isn't then let's address it in a separate PR.

packages/shared/components/popups/send/SendConfirmationPopup.svelte
```diff
@@ -173,19 +214,45 @@
...
+                asset={transactionDetails.asset}
+                {storageDeposit}
+                subject={recipient}
+                rawAmount={transactionDetails.rawAmount}
+                tag={transactionDetails.tag}
+                unit={transactionDetails.unit}
+                {isInternal}
+                {surplus}
+                type={ActivityType.Transaction}
+                metadata={transactionDetails.metadata}
```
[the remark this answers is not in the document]
MarkNerdi: this component is used in other places as well, where we don't have this `transctionDetails` store object. i can simplify  a few of those fields, but many will remain. 

A solution would be to do it like we did previously in this file by creating a bundling object (see `ActivityDetailsPopup` -> `getActivityDetails()`) and deconstruct that when passing the single props to the component.
I personally think its cleaner passing the props explicitly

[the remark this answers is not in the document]
MarkNerdi: this component is used in other places as well, where we don't have this transctionDetails store object. i can simplify a few of those fields, but many will remain.

A solution would be to do it like we did previously in this file by creating a bundling object (see ActivityDetailsPopup -> getActivityDetails()) and deconstruct that when passing the single props to the component.
I personally think its cleaner passing the props explicitly

packages/shared/components/popups/send/SendConfirmationPopup.svelte
```diff
@@ -98,37 +97,80 @@
...
+            transactionDetails.nftId,
+            transactionDetails.immutableFeatures,
+            transactionDetails.metadata,
+            transactionDetails.tag,
+            recipientAddress,
+            $selectedAccount.depositAddress,
+            storageDeposit,
+            giftStorageDeposit
+        )
+        preparedOutput = await $selectedAccount.buildNftOutput(outputData)
```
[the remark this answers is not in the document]
MarkNerdi: true, didnt see that I can use the nftId inside that, that makes everything a lot easier

packages/shared/lib/core/wallet/utils/buildNftOutputData.ts
MarkNerdi: i agree that a lot of arguments are hard to read, but i'm not sure how you would simplify this?

now in this case we don't need it anymore, because the function is now unused, but how would you minimize the args?

packages/shared/components/molecules/NftDetails.svelte
```diff
@@ -63,15 +66,25 @@
...
+            metadata: {
+                data: metadata,
+                tooltipText: localize(localePrefix + 'metadata'),
+            },
+        }),
+        ...(tag && {
+            tag: {
+                data: tag,
+                tooltipText: localize(localePrefix + 'tag'),
+            },
```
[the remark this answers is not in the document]
MarkNerdi: i think for most fields its unnecassary. Tag and metadata can be copyable, but fields like storagedeposit, tx date or expiration date shouldnt be copyable imo

packages/shared/components/popups/send/SendConfirmationPopup.svelte
```diff
@@ -226,30 +210,24 @@
     <div class="w-full flex-col space-y-2">
         {#if transactionDetails.type === NewTransactionType.TokenTransfer}
             <TransactionDetails
-                asset={transactionDetails.asset}
+                {...transactionDetails}
                 {storageDeposit}
                 subject={recipient}
```
Tuditi, line 190: This is also part of transactionDetails (-> `transactionDetails.recipient`)

packages/shared/components/popups/send/SendFormPopup.svelte
```diff
@@ -33,24 +32,17 @@
 
     let rawAmount: string
     let asset: IAsset
-    let metadata: string
-    let recipient: Subject
-    let tag: string
     let unit: string
 
     const transactionDetail = get(newTransactionDetails)
```
Tuditi, line 37: Maybe use `transactionDetails` for consistency?

packages/shared/lib/core/wallet/utils/getOutputOptions.ts
```diff
@@ -25,6 +26,19 @@ export function getOutputOptions(
...
+    const assets: Assets = {}
+    if (nativeTokenId) {
+        assets.nativeTokens = [
+            {
+                id: nativeTokenId,
+                amount: '0x' + bigAmount.toString(16),
+            },
+        ]
+    }
+    if (nftId) {
```
Tuditi, line 38: At the moment wallet.rs doesn't allow to set both `nativeTokenId` & `nftId`. If there is an `nftId`, it should take precedence over the `nativeTokenId`.

Tuditi: https://github.com/iotaledger/wallet.rs/blob/feat/issuer-and-sender-to-output/src/account/operations/transaction/prepare_output.rs#L37-L39

Tuditi (commented):
Added some extra comments. :smiling_imp:

packages/shared/lib/core/wallet/types/new-transaction-details.type.ts
```diff
@@ -1,13 +1,8 @@
...
 
-export type NewTransactionDetails = NewNftTransactionDetails | NewTokenTransactionDetails
+export type NewTransactionDetails = NewBaseTransactionDetails & (NewNftTransactionDetails | NewTokenTransactionDetails)
 
-export type NewTokenTransactionDetails = {
-    type: NewTransactionType.TokenTransfer
-    rawAmount: string
-    asset: IAsset
-    unit: string
+export type NewBaseTransactionDetails = {
```
Tuditi, line 5: I wouldn't export this & extend respective types on l.16 + l.23 with it

Tuditi (approved):
Pfew, thanks for this effort! LGTFM! :D

PR #5091, opened 2022-11-02
refactor: improve profile object migration

maxwellmattryan:
## Summary

The migration logic was incorrect, causing certain properties to be removed.

## Changelog

```
- Adjusted migration logic for profiles
- Move it to per-app instead of per-profile
```

## Relevant Issues

...

### Platforms

- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

### Instructions

Test creating profiles, logging in / out of new and existing profiles, restarting the app, etc.

## Checklist


nicole-obrien (changes requested):
Migration tests are failing

nicole-obrien (dismissed):
Additional testing it does not remove properties from old objects that are not in the new format.

packages/shared/lib/core/utils/object.ts
```diff
@@ -10,30 +10,33 @@
 export function migrateObjects<T>(oldObj: T, newObj: T): T {
     /* eslint-disable @typescript-eslint/no-explicit-any */
     const _helper = (curObj: any, oldObj: any, newObj: any): any => {
-        // Iterate through each key of a new object...
-        for (const k in newObj) {
+        // Iterate through each key of a new AND old objects...
+        for (const key in { ...oldObj, ...newObj }) {
```
MarkNerdi, line 14: with this change, properties which are in `oldObj` but not in `newObj` are not deleted

packages/shared/lib/core/utils/object.ts
```diff
@@ -10,30 +10,33 @@
...
+                if (typeof newObj[key] === 'object' && newObj[key] !== undefined && newObj[key] !== null) {
                     // If corresponding value is actually an array (b/c an "Array" is an object in JS) then...
-                    if (Array.isArray(newObj[k])) {
+                    if (Array.isArray(newObj[key])) {
                         // Create key-value pair with array from new object or old object if it already exists
-                        curObj[k] = Array.isArray(oldObj[k]) ? oldObj[k] : newObj[k]
+                        curObj[key] = Array.isArray(oldObj[key]) ? oldObj[key] : newObj[key]
+                        // Else if corresponding value is actually a date
+                    } else if (typeof newObj[key]['getMonth'] === 'function') {
+                        curObj[key] = newObj[key]
```
MarkNerdi, line 25: if we have a date in the `oldObj`, it gets always overridden with the `newObj`s date

packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
```diff
@@ -1,13 +1,14 @@
...
+import { sanitizeOnboardingProfile } from './sanitizeOnboardingProfile'
 
 export function convertOnboardingProfileToPersistedProfile(
     onboardingProfile: Partial<IOnboardingProfile>
 ): IPersistedProfile {
     return {
-        ...DEFAULT_ACTIVE_PROFILE_VALUE,
-        ...onboardingProfile,
+        ...DEFAULT_PERSISTED_PROFILE_OBJECT,
+        ...sanitizeOnboardingProfile(onboardingProfile),
```
MarkNerdi, line 12: if we pass in here an `IOnboardingProfile` and only sanitize `importFile`, `importFilePath`, `mnemonic` and `strongholdPassword`, the resulting object will contain more fields than the return type of `convertOnboardingProfileToPersistedProfile`, e.g. `protectionType`, ` hasStoreMnemonic` etc. 

Either we need to update the return type or make sure that only the relevant fields are returned

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
```diff
@@ -0,0 +1,69 @@
...
+    for (const profile of _profiles) {
+        migratePersistedProfile(profile)
+    }
+}
+
+function migratePersistedProfile(persistedProfile: IPersistedProfile): void {
+    let migratedPersistedProfile = persistedProfile
+    let migrationVersion = get(currentProfileVersion)
+    for (migrationVersion; migrationVersion < PROFILE_VERSION; migrationVersion++) {
+        migratedPersistedProfile = persistedProfileMigrationsMap?.[migrationVersion]?.(migratedPersistedProfile)
```
MarkNerdi, line 30: we can remove the first `?` because it is always defined

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
```diff
@@ -0,0 +1,69 @@
...
+        'hasVisitedDashboard',
+        'lastUsedAccountIndex',
+        'clientOptions',
+    ]
+
+    keysToKeep.forEach((key) => {
+        const existingValue = existingProfile?.[key]
+        if (existingValue) {
+            newProfile[key] = existingValue
+        }
```
MarkNerdi, line 65: i think we need to remove that check
if the old profile contains a `someCond: false` , `someNum: 0` or  `settings: null/undefined`, we still want that in the new object

MarkNerdi (commented):
nice structure and works good. 
Pointed out a few things to make sure the objects are as we expect them to be

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
nicole-obrien: Which question mark? If someone it on a version or we skip a migration for a version then it will break truing to call this function . I think actually we should replace it with
```suggestion
        migratedPersistedProfile = persistedProfileMigrationsMap?.[migrationVersion]?.(migratedPersistedProfile) ?? migratedPersistedProfile 
```

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
nicole-obrien: ahh so it should be === undefined?

packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
nicole-obrien: Updated

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
nicole-obrien: Updated

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
nicole-obrien: Updated

packages/shared/lib/core/profile/actions/active-profile/checkAndMigrateProfiles.ts
```diff
@@ -0,0 +1,70 @@
...
+        'accountMetadata',
+        'isDeveloperProfile',
+        'hasVisitedDashboard',
+        'lastUsedAccountIndex',
+        'clientOptions',
+    ]
+
+    keysToKeep.forEach((key) => {
+        const existingValue = existingProfile?.[key]
+        if (existingValue !== undefined) {
```
MarkNerdi, line 64: i think we dont need a check at all, because we want each of the `keysToKeep` to be set, right?

so if `existingProfile = {settings: undefined, clientOptions: undefined}`, we also want `newProfile`  to contain `settings: undefined, clientOptions: undefined`

packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
```diff
@@ -1,13 +1,28 @@
...
 import { IOnboardingProfile } from '../interfaces'
 
 export function convertOnboardingProfileToPersistedProfile(
     onboardingProfile: Partial<IOnboardingProfile>
 ): IPersistedProfile {
     return {
-        ...DEFAULT_ACTIVE_PROFILE_VALUE,
-        ...onboardingProfile,
+        ...DEFAULT_PERSISTED_PROFILE_OBJECT,
+        ...(onboardingProfile?.id && { id: onboardingProfile.id }),
```
MarkNerdi, line 11: we dont need these optional properties, because e.g. an `IPersistedProfile` should always include an `id` property. So if you just add `id:  onboardingProfile.id` even if `onboardingProfile` has no `id`, the `IPersistedProfile` will contain that property

nicole-obrien: Yeah but I don't want to override it if with undefined

MarkNerdi: ah i just saw that you would override the default profile, ignore this comment

MarkNerdi (approved):
LGTM

PR #5354, opened 2022-12-05
fix: selector input unwanted readonly

jeeanribeiro:
## Summary
`readonly` prop should have an initial value of null or undefined, because svelte is adding the attribute in the HTML, and the HTML does not understand the `readonly` attribute with false or true value, just understand the absence of `readonly` prop

## Changelog
```
- Sets the initial value of readonly prop in SelectorInput to null
```

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android
  
## Checklist

### DIFF
--- a/packages/shared/components/inputs/AssetAmountInput.svelte
+++ b/packages/shared/components/inputs/AssetAmountInput.svelte
@@ -20,7 +20,7 @@
     export let rawAmount: string = undefined
     export let unit: string = undefined
     export let containsSlider: boolean = false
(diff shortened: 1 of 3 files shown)


Tuditi (commented):
I added a commit to fix it also on NftInput and AssetAmountInput

MarkNerdi (approved):
LGTM

PR #5440, opened 2022-12-17
feat: show vote percentages

### LINKED ISSUE #5429: [Task]: Show vote percentages
### Task description

When navigating to a  proposal the already answers that a user voted on should be displayed according to the design.

### Requirements

TBD

### Creation checklist

jeeanribeiro:
## Summary
Change the hasVoted state if the proposal has been voted by the selected account or not

## Changelog

```
- Adds isVotingForProposal function
- Refactors isVotingForSelectedProposalFunction
- Removes 'hasVoted' prop from the Proposal interface
- Adds the necessary logic to display if the proposal has really been voted
```

## Relevant Issues
closes #5429 

### Platforms
- __Desktop__
  - [ ] MacOS
  - [x] Linux
  - [ ] Windows
- __Mobile__
  - [ ] iOS
  - [ ] Android

## Checklist


Tuditi:
I moved this to draft, since this feature should also include the actual votes that have been voted for.

image

packages/shared/components/ProposalQuestion.svelte
```diff
@@ -33,14 +37,15 @@
             <Icon icon={IconEnum.ChevronDown} classes="text-gray-500" />
         </div>
     </div>
-    <proposal-answers class:mt-4={isOpened} class="space-y-2">
-        {#each question.answers as answer, answerIndex}
+    <proposal-answers class:mt-4={isOpened || voteValue} class="space-y-2">
```
maxwellmattryan, line 40: I think `hasVoteValue` might be better because if the user has abstained then `0` would evaluate to false.

packages/shared/components/organisms/ProposalCard.svelte
```diff
@@ -9,10 +9,18 @@
...
     export let proposal: IProposal
 
+    let hasVoted = false
+
     $: proposalState = $proposalsState[proposal?.id]
+    $: void setHasVoted()
+
+    async function setHasVoted(): Promise<void> {
+        hasVoted = await isVotingForProposal(proposal?.id)
+    }
```
maxwellmattryan, line 23: This could be inside of an `onMount`:
```typescript
let hasVoted: boolean

async function onMountHelper(): Promise<void> {
    hasVoted = await isVotingForProposal(proposal?.id)
}

onMount(() => void onMountHelper())
```

maxwellmattryan (changes requested):
Looks good! Just requested a few minor changes :+1:

packages/shared/components/ProposalQuestion.svelte
Tuditi: The problem is that there is no difference between a user abstaining the vote or never having voted for it. Both would result in `0`

Tuditi (approved):
LGTM!

packages/shared/components/ProposalQuestion.svelte
```diff
@@ -3,18 +3,36 @@
     import { Icon as IconEnum } from '@auxiliary/icon'
     import type { Question } from '@iota/wallet'
 
-    export let question: Question
+    export let currentVote: Question = undefined
     export let index: number = undefined
```
Tuditi, line 7: Maybe it's worth renaming this to questionIndex?

packages/shared/components/ProposalQuestion.svelte
```diff
@@ -3,18 +3,36 @@
...
+        percentages = answers?.map((currentAnswer) => {
+            const answerAccumulated = currentVote?.answers?.find(
+                (answer) => answer.value === currentAnswer.value
+            )?.accumulated
+            const divisionResult = answerAccumulated / totalAccumulated
+            return Number.isNaN(divisionResult) ? '0%' : `${Math.round(divisionResult * 100)}%`
+        })
+    }
+
+    function handleAnswerClick(answer: number): void {
```
Tuditi, line 31:
```suggestion
    function handleAnswerClick(answerValue: number): void {
```

Tuditi (changes requested):
Looks good, one minor change though:

Some changes required:

1) Percentages should be of a color gray-700
2) There should be placeholder for a tooltip icon and the percentage should always be left of that.
image

Tuditi (changes requested):
Could you make it so that the percentage is always aligned?
image

Tuditi (approved):
LGTM! :fire:

PR #5558, opened 2023-01-11
feat: automatic proposal querying

maxwellmattryan:
## Summary

Official proposals should be automatically registered to the user's Governance dashboard.

## Changelog

```
- Add registerProposalsFromNode function called in login
- Add registerProposalsFromPrimaryNode
- Cleanup imports in login file
- Register proposals when new primary node is selected
- Set primary node during onboarding
```

### Platforms

- __Desktop__
	- [ ] MacOS
	- [x] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions

Create new profiles connected to node with voting events. Log into the profile more than once.

## Checklist


maxwellmattryan:
@Tuditi One change I think can be made is that we don't register a proposal if it has already ended, but some things worth noting:
- Some users may want to be able to see all proposals, even if they've ended, so that they can see the results of the vote
- Checking the status means adding an extra API call per proposal ID that we get. We could suggest for the node team to change this API endpoint so that it returns the entire event rather just the ID, which would be sensible since it's unlikely that there will be many many events on a node anyway (but who knows?)

My thoughts are that we should just keep it simple for now and register everything, then we can see based on beta testing how users feel about having every single event registered.

packages/shared/lib/contexts/governance/utils/getVotingEventIds.ts
```diff
@@ -0,0 +1,9 @@
+import { get } from 'svelte/store'
+
+import { ParticipationEventId, ParticipationEventType } from '@iota/wallet/out/types'
+
+import { selectedAccount } from '@core/account/stores'
+
+export function getVotingEventIds(): Promise<ParticipationEventId[]> {
```
nicole-obrien, line 7:
```suggestion
export function getVotingParticipationEventIds(): Promise<ParticipationEventId[]> {
```

packages/shared/lib/contexts/governance/utils/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './getNumberOfVotedProposals'
 export * from './getNumberOfVotingProposals'
 export * from './getParticipationsForProposal'
 export * from './getPercentagesFromAnswerStatuses'
+export * from './getVotingEventIds'
```
nicole-obrien, line 11:
```suggestion
export * from './getVotingParticipationEventIds'
```

packages/shared/lib/contexts/governance/actions/registerProposalsFromPrimaryNode.ts
```diff
@@ -0,0 +1,34 @@
+import { get } from 'svelte/store'
+
+import { selectedAccount } from '@core/account/stores'
+import { activeProfile } from '@core/profile/stores'
+
+import { getVotingEventIds, isProposalAlreadyAddedForSelectedAccount } from '../utils'
+
+import { registerParticipationEvent } from './registerParticipationEvent'
+
+export async function registerProposalsFromPrimaryNode(): Promise<void> {
```
nicole-obrien, line 10: Can this be more generic by just passing in a node and then we can decide what nodes to use where we call this. i.e. we don't usually have a primary node

packages/shared/lib/contexts/governance/actions/registerProposalsFromPrimaryNode.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!proposalIds || proposalIds.length === 0) {
+        return
+    }
+
+    const clientOptions = get(activeProfile)?.clientOptions
+    if (!clientOptions) {
+        throw new Error('Unable to retrieve profile client options')
+    }
+
+    const primaryNode = clientOptions?.primaryNode ?? clientOptions?.nodes[0]
```
nicole-obrien, line 21: We should just pass in the nodes, and we can call this in a wrapper function for all the nodes the user has.

nicole-obrien (changes requested):
Left some comments, I recently cleaned up all our api requests so its not taking as long to load so we should be able to pull in all events from our nodes at login. 

We could also register the events for a node when we add it.

I also need to double check but the questions where not showing on the proposal details page 🤷‍♀️

nicole-obrien:
> Checking the status means adding an extra API call per proposal ID that we get. We could suggest for the node team to change this API endpoint so that it returns the entire event rather just the ID, which would be sensible since it's unlikely that there will be many many events on a node anyway (but who knows?)

I have already changed this, we a don't need any api call check the status of an event it can be calculated by the current milestone.

packages/shared/lib/contexts/governance/actions/registerProposalsFromPrimaryNode.ts
```diff
@@ -0,0 +1,19 @@
+import { get } from 'svelte/store'
+
+import { activeProfile } from '@core/profile/stores'
+
+import { registerProposalsFromNode } from './registerProposalsFromNode'
+
+export async function registerProposalsFromPrimaryNode(): Promise<void> {
```
nicole-obrien, line 7: If we don't have a primary node we do not need to throw an error, just dont attempt to register any proposals for that node.

packages/shared/lib/contexts/governance/actions/registerProposalsFromNode.ts
```diff
@@ -0,0 +1,28 @@
...
+import { INode } from '@core/network/interfaces'
+
+import { getVotingParticipationEventIds, isProposalAlreadyAddedForSelectedAccount } from '../utils'
+
+import { registerParticipationEvent } from './registerParticipationEvent'
+
+export async function registerProposalsFromNode(node: INode): Promise<void> {
+    if (!node) {
+        throw new Error('Unable to determine node')
+    }
```
Tuditi, line 13: This error is redundant if you follow Nicole's suggestion.

Tuditi (approved):
Looks good, I left a comment wrt Nicole's requested changes.
Additionally, we'll need a way to store removed event IDs, since removed official events get added back on login.

Tuditi (changes requested):
Tests are failing

maxwellmattryan:
@Tuditi

maxwellmattryan:
@Tuditi,

> Additionally, we'll need a way to store removed event IDs, since removed official events get added back on login.

Agreed, but per the call we had today let's do this as a separate task so we can get a release out.

nicole-obrien (approved):
LGTM!

PR #5573, opened 2023-01-13
fix: proposal removal logic and UI

jeeanribeiro:
## Summary
Fixes the proposal removal logic to only allow removing a proposal when there is not any other accounts voting for the same proposal. Adds isBusy state when this condition is being calculated.

## Changelog

```
- Adds isBusy state in ProposalDetailsMenu
- Adds async to isVotingForProposal functions
- Fix proposal removal logic
```

### Platforms
- __Desktop__
	- [ ] MacOS
	- [ ] Linux
	- [x] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

## Checklist


packages/shared/lib/contexts/governance/utils/isVotingForSelectedProposal.ts
```diff
@@ -3,7 +3,7 @@ import { get } from 'svelte/store'
 import { selectedProposal } from '../stores'
 import { isVotingForProposal } from './isVotingForProposal'
 
-export function isVotingForSelectedProposal(): boolean {
+export async function isVotingForSelectedProposal(accountIndex?: number): Promise<boolean> {
     const proposalId = get(selectedProposal)?.id
-    return isVotingForProposal(proposalId)
+    return isVotingForProposal(proposalId, accountIndex)
```
maxwellmattryan, line 8: We may want to use this function in a `try/catch` block, so we need to `return await ...` this line, i.e. 
```suggestion
    return await isVotingForProposal(proposalId, accountIndex)
```

See more [here](https://stackoverflow.com/a/42750371/11150203) :+1:

packages/shared/locales/en.json
```diff
@@ -1950,6 +1950,9 @@
             "description": "Optional parameter: A description of the NFT",
             "attributes": "Optional parameter: An array of traits and values that define attributes of the NFT",
             "uri": "To create a URI using custom media, first upload your file to IPFS via a storage service (e.g. https://nft.storage/)"
+        },
+        "governance": {
+            "removeProposalWarning": "You must stop voting for this proposal in all accounts before removing it."
```
maxwellmattryan, line 1955: Not a requested change, but this makes me think we should move the proposal registration to be account-based rather than profile-based ASAP.

packages/shared/lib/contexts/governance/utils/isVotingForProposal.ts
```diff
@@ -2,15 +2,13 @@ import { get } from 'svelte/store'
...
-    const overview = get(participationOverview)
-    const participations = overview?.participations?.[proposalId]
-    if (participations) {
-        const participationOutputs: TrackedParticipationOverview[] = Object.values(participations)
-        return participationOutputs.some((output) => output?.endMilestoneIndex === 0)
-    } else {
-        return false
-    }
+export async function isVotingForProposal(proposalId: string, accountIndex?: number): Promise<boolean> {
+    const _accountIndex = accountIndex ?? get(selectedAccount)?.index
```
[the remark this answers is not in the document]
jeeanribeiro: This is on purpose, immutability is good practice and in this case it is a must because we shouldn't change the function parameters

packages/shared/lib/contexts/governance/utils/isVotingForSelectedProposal.ts
jeeanribeiro: Nice StackOverflow link!

What I've understood is that we could return a Promise resolve even after the code reaches the catch block, so where this function is called we could not try-catch it, it would always be resolved, is that it?

If this is for the sole purpose of calling the "isVotingForSelectedProposal" function inside a try-catch block, then I think this change in the code is not necessary

maxwellmattryan: The try/catch block doesn't have to be inside the called function, it's a matter of how the caller function uses it. If the caller function uses this function inside its own try/catch block (and it doesn't have the `await`) then it won't trigger the catch, so I do think it is necessary given that we use try/catch blocks often.

maxwellmattryan:
Have tested it out and have some feedback:

- With funds actively voted on wallet A and no funds on B, whenever I switch accounts from B to A, there is a small amount of time where the "Remove proposal" menu item is disabled with the spinner. Soon after it disappears and is replaced with the "Stop voting" menu item. The problem seems to be with the loading speed of the participation data when switching accounts, so we could load participation data for all accounts on login but only poll data for the selected account, just so we have some data when switching. This is a heftier change and would require us to re-structure participation data so that it includes all accounts. It would potentially mean even more API requests on login, although really not sure that many users would be voting with multiple accounts. Not sure about this though... :thinking: 
- With funds actively voted on wallet A and no funds on B, whenever I switch accounts from B to A, there is a small amount of time when I can click on "Remove proposal" and actually remove it for all accounts despite A currently voting for it. It's the same problem of the participation data not loading fast enough... although this could be solved if we make the proposal registration per-account, so maybe we should probably expedite making that change.

jeeanribeiro:
> Have tested it out and have some feedback:
> 
> * With funds actively voted on wallet A and no funds on B, whenever I switch accounts from B to A, there is a small amount of time where the "Remove proposal" menu item is disabled with the spinner. Soon after it disappears and is replaced with the "Stop voting" menu item. The problem seems to be with the loading speed of the participation data when switching accounts, so we could load participation data for all accounts on login but only poll data for the selected account, just so we have some data when switching. This is a heftier change and would require us to re-structure participation data so that it includes all accounts. It would potentially mean even more API requests on login, although really not sure that many users would be voting with multiple accounts. Not sure about this though... 🤔
> * With funds actively voted on wallet A and no funds on B, whenever I switch accounts from B to A, there is a small amount of time when I can click on "Remove proposal" and actually remove it for all accounts despite A currently voting for it. It's the same problem of the participation data not loading fast enough... although this could be solved if we make the proposal registration per-account, so maybe we should probably expedite making that change.

Nice catch!!!
While we don't have the proposal registration per account I've changed the solution a little bit so we avoid these unwanted scenarios

packages/shared/lib/contexts/governance/utils/isVotingForSelectedProposal.ts
jeeanribeiro: I made this example to test the try/catch block without return await: https://playcode.io/1066291
You can change the `VALIDATED_STRING` value to test the resolve/reject

maxwellmattryan:
Looking really good! Only piece of feedback is that the spinner looks a little bit strange in that the icon is smaller than and not in the same position as the menu item buttons.

maxwellmattryan (changes requested):
One minor requested change to adjust the UI you recently added for the spinner in the proposal menu while the data is being fetched :pray:

maxwellmattryan (approved):
LGTM :fire:

Issue #5265, opened 2022-11-21
[Task]: Add ability to manage voting power while voting


maxwellmattryan:
### Task description

Users will need to be able to manage their voting power from the main Governance dashboard, i.e. increasing or decreasing the amount that they are able to vote with. This has to be compatible with the fact that they are voting at the moment.

### Requirements

_TBD_

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #5593, opened 2023-01-16
feat: manage voting power while voting

Tuditi:
## Summary

Allow managing the voting power while voting.

closes: https://github.com/iotaledger/firefly/issues/5265

## Changelog

```
- make accountIndex obligatory to vote()
- add revotePopup
```

### Platforms

- __Desktop__
	- [ ] MacOS
	- [ ] Linux
	- [ ] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

### Instructions

...

## Checklist


maxwellmattryan (commented):
Looks great! Just a few minor suggested changes :pray:

maxwellmattryan:
Some feedback after testing:

- The slider is greyed out once the manage voting power popup is re-opened, but...
    - The amount input is still enabled
    - The slider shows the previous amount rather than the newly selected amount
    - The unit dropdown is still enabled
- If the user has the same voting power amount selected, we should disable the "Confirm" button (in the manage voting power popup)
- The text hint in the revote popup should (IMO) be an "info" one rather than "success"
- The revote popup should say "Revote" rather than "Confirm" (IMO)
- As discussed in our review call, we'll likely need to prevent closing of the revote popup, forcing the user to take an action (either re-vote or completely unvote)
- Not necessarily a problem, but just FYI I can navigate to settings (via the menu bar) while the manage voting power popup is confirming a tx
- If I click out of the manage voting power popup while the transaction is being confirmed (after it's broadcasted), then the revote popup isn't opened until I click the manage voting power button again

packages/shared/lib/contexts/governance/actions/setVotingPower.ts
```diff
@@ -1,29 +1,37 @@
-import { selectedAccount, updateSelectedAccount } from '@core/account'
+import { get } from 'svelte/store'
+import { Transaction } from '@iota/wallet/out/types'
+import { hasToRevote } from '@contexts/governance/stores'
```
maxwellmattryan, line 3: This should be a relative import

packages/shared/lib/contexts/governance/utils/isSelectedAccountVoting.ts
```diff
@@ -0,0 +1,7 @@
+import { get } from 'svelte/store'
+import { DEFAULT_PARTICIPATION_OVERVIEW } from '../constants'
+import { participationOverview } from '../stores'
+
+export function isSelectedAccountVoting(): boolean {
+    return get(participationOverview) === DEFAULT_PARTICIPATION_OVERVIEW
```
maxwellmattryan, line 6: Why would this comparison indicate `isSelectedAccountVoting`?

maxwellmattryan (changes requested):
Lots of good work - just have a few questions / comments.

packages/shared/components/popups/VoteForProposalPopup.svelte
```diff
@@ -26,28 +23,11 @@
...
                 closePopup()
             })
         } catch (err) {
-            if ($activeProfile.type === ProfileType.Ledger) {
-                handleLedgerError(err)
-            } else {
-                handleError(err)
-            }
-            updateSelectedAccount({ isTransferring: false })
+            console.error(err)
```
maxwellmattryan, line 30: This should be `handleError`

maxwellmattryan (changes requested):
Another requested change

maxwellmattryan:
Something strange that's happening whenever I modify the voting power and it stays within the manage voting power popup - the maximum slider value changes to whatever my available balance plus the difference between the old and new voting power. So I've got 1,000 SMR total, going from voting power 300 to 400, then the slider max value becomes 1,100 SMR.

packages/shared/lib/contexts/governance/actions/vote.ts
```diff
@@ -0,0 +1,32 @@
...
+            type: 'success',
+            message: localize('notifications.vote.success'),
+            alert: true,
+        })
+    } catch (err) {
+        const _activeProfile = get(activeProfile)
+        if (_activeProfile.type === ProfileType.Ledger) {
+            handleLedgerError(err?.error)
+        }
+        return Promise.reject(err)
```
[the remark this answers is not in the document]
Tuditi: Yeah, we can and it simplifies the code a bit, so I added the changes.

packages/shared/lib/contexts/governance/utils/isSelectedAccountVoting.ts
Tuditi: My bad, this is wrong! I fixed it. :sweat_smile:

maxwellmattryan:
Two pieces of feedback:

- If the user isn't voting and I modify the voting power, it feels strange that the popup doesn't close whenever the operation is complete (I don't mean when the tx inclusion state is confirmed).
- I'm able to somehow cast "votes" for a proposal with 0 SMR voting power. I'd argue it's outside the scope of this PR, but I'm able to see the "Vote for proposal popup", and weirdly by clicking on the key-value box it takes me to the password input popup. When I actually try to vote I get an error `No unspent voting output found`. This was after I set my voting power to zero while voting.
- Clicking on the key-value box in the "Vote for proposal" popup allows me to vote (outside the scope of this PR)

maxwellmattryan:
The code generally looks good and the app works as expected :fire: 

I think there is some room for polish and some issues that are out of the scope of the PR, but is ready to merge IMO.

maxwellmattryan (approved):

PR #5655, opened 2023-01-23
feat: governance polishing

jeeanribeiro:
## Summary
UI/UX polishing, all fixes/enhancements are described in the commit history and in the related issue

### Platforms
- __Desktop__
	- [ ] MacOS
	- [ ] Linux
	- [x] Windows
- __Mobile__
	- [ ] iOS
	- [ ] Android

## Checklist


packages/desktop/views/dashboard/governance/views/DetailsView.svelte
```diff
@@ -122,6 +124,14 @@
...
             selectedAnswerValues[questionIndex] = answerValue
         }
+
+        openedQuestionIndex = selectedAnswerValues.length === questionIndex + 1 ? null : questionIndex + 1
+        const selectedQuestionElement: HTMLElement = proposalQuestions?.querySelector(
+            'proposal-question:nth-child(' + (openedQuestionIndex + 1) + ')'
+        )
+        setTimeout(() => {
+            proposalQuestions.scrollTo({ top: selectedQuestionElement?.offsetTop, behavior: 'smooth' })
+        }, 500)
```
Tuditi, line 134: Works like a charm, but it feels quite slow. Maybe 100/200 is a better value? Maybe we should extract this to a constant, so we have the same delay wherever user interaction is required?

packages/shared/components/organisms/ProposalCard.svelte
```diff
@@ -32,10 +32,12 @@
 
 <proposal-card
     on:click={handleProposalClick}
-    class="flex flex-col p-6 border border-solid border-gray-200 dark:border-transparent rounded-xl cursor-pointer h-32
-    {proposal.status === ProposalStatus.Ended ? 'bg-transparent' : 'bg-white dark:bg-gray-850'}"
+    on:keydown={(e) => e.key === 'Enter' && handleProposalClick()}
```
Tuditi, line 35: This doesn't work for me

packages/shared/components/organisms/ProposalCard.svelte
```diff
@@ -32,10 +32,12 @@
...
+    class="flex flex-col p-6 border border-solid border-gray-200 dark:border-transparent rounded-xl cursor-pointer h-fit shadow-elevation-1 hover:shadow-inner focus:shadow-inner
```
Tuditi, line 36: I saw that this has been added to the task, but I'm not aware that we needed shadows. I asked design team to confirm this? I really dislike the hover shadow and we don't have it anywhere else as far as I'm aware

packages/shared/components/popups/AddProposalPopup.svelte
```diff
@@ -64,12 +66,15 @@
...
 
     async function registerParticipationWrapper(auth?: Auth): Promise<void> {
-        await registerParticipationEvent(eventId, [{ url: nodeUrl, auth }])
+        const event = await registerParticipationEvent(eventId, [{ url: nodeUrl, auth }])
         showAppNotification({
             type: 'success',
             message: localize('views.governance.proposals.successRegister'),
             alert: true,
         })
+        const proposal = await createProposalFromEvent(event)
```
Tuditi, line 75: Don't need to await here.
```suggestion
        const proposal = createProposalFromEvent(event)
```

packages/shared/locales/en.json
```diff
@@ -577,15 +577,18 @@
...
             },
             "proposals": {
                 "title": "Proposals",
                 "voted": "Voted",
-                "successRegister": "Proposal successfully registered",
-                "successRemove": "Proposal successfully removed"
+                "successAdd": "Proposal successfully added",
+                "successRemove": "Proposal successfully removed",
+                "emptyTitle": "No proposals",
+                "emptyDescription": "Add a proposal now!"
```
Tuditi, line 591: Don't kill me if I don't :rofl:

Tuditi: I'll ask design to get something smoother here :D

Tuditi (changes requested):
Looks good overall, some minor changes need to be done and then it's ready to go!

packages/shared/locales/en.json
jeeanribeiro: 😂 Just adapted from collectibles

Tuditi: We can remove it, since it's the same message as the button.

packages/shared/components/organisms/ProposalCard.svelte
jeeanribeiro: I will remove the hover effect but I will keep the focus one, we need some visual guiding to the keyboard tabbing

packages/shared/components/organisms/ProposalCard.svelte
jeeanribeiro: It it working here, you need to tab until the focus is on one of the proposal cards and then press `Enter`

packages/desktop/views/dashboard/governance/views/DetailsView.svelte
jeeanribeiro: I think it does not need to be a constant, it could be different for other components, I put 500ms on purpose, so the user has visual feedback that the clicked answer was selected, but I will reduce the time if it is feeling slow

packages/shared/locales/en.json
jeeanribeiro: Removed the "emptyDescription"

Tuditi (changes requested):
One thing I noticed is that the margins for selected votes have changed again. That's the only thing I could find that needs changing

image

Tuditi (approved):
LGTM! :fire:

PR #5702, opened 2023-01-30
enhancement: adjust default proposal display order

maxwellmattryan:
## Summary

The user should see a default ordering that is more sensible for proposals.

## Changelog

```
- Added constants
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

Check default sorting by registering multiple proposals in different phases, including ones with the same phases.

## Checklist


packages/shared/lib/contexts/governance/utils/sortProposals.ts
```diff
@@ -21,16 +22,42 @@ export function sortProposals(proposals: IProposal[], filter: ProposalFilter): I
     return proposals?.sort((proposal1, proposal2) => orderFunction(proposal1, proposal2, isAscending)) ?? []
 }
 
+function sortByPhaseAndMilestone(proposal1: IProposal, proposal2: IProposal, asc: boolean): number {
+    return sortByPhase(proposal1, proposal2, asc) || sortByMilestone(proposal1, proposal2, asc)
```
MarkNerdi, line 26: as far as i can see, the second part of the condition will never be called, because  `sortByPhase` will either return 1 or -1. It should return 0 if the phase is the same

packages/shared/lib/contexts/governance/utils/sortProposals.ts
```diff
@@ -21,16 +22,42 @@ export function sortProposals(proposals: IProposal[], filter: ProposalFilter): I
...
+        case ProposalStatus.Commencing:
+            proposal1Milestone = proposal1.milestones.holding
+            proposal2Milestone = proposal2.milestones.holding
+            break
+        case ProposalStatus.Holding:
+        case ProposalStatus.Ended:
+            proposal1Milestone = proposal1.milestones.ended
+            proposal2Milestone = proposal2.milestones.ended
+            break
     }
```
MarkNerdi, line 60:
```suggestion
            const proposal1Milestone = proposal1.milestones[proposal1.status]
            const proposal2Milestone = proposal2.milestones[proposal2.status]
```

MarkNerdi (changes requested):
Nice cleanup!
left 2 request changes

packages/shared/lib/contexts/governance/utils/sortProposals.ts
maxwellmattryan: Exactly, so if it is `0` it evaluates to false, which means it will get sorted on the secondary criteria which is the milestone.

packages/shared/lib/contexts/governance/utils/sortProposals.ts
maxwellmattryan: We can do this, but it would mean that the milestones are sorted on when the phase started, but IMO it's more useful to sort based off when the current phase ends / next phase starts. If have proposal A that started way long ago and ends in a year and proposal B that started after A but ends in a day, I would think the user wants to see B first?

MarkNerdi: aaaah sorry i didnt see that! i saw that on the last case we could do that, but didnt saw that on the other ones we look at different milestones

MarkNerdi (approved):
LGTM

PR #5742, opened 2023-02-01
fix: busy states during governance transactions

### LINKED ISSUE #5726: [Task]: Disable voting button until governance transaction is included on the Tangle
### Task description

At the moment, it's possible to vote twice in quick succession on the same option. The issue is that we rely entirely on `isTransferring` to show the busy state. For governance transaction, we should store the last voting transaction ID and remove it from the store if it has been included. 

Additionally we should update the participation overview when we sent a vote.

### Requirements

TBD

### Creation checklist

nicole-obrien:
Co-authored-by: Tuditi <Tuditi@users.noreply.github.com>

## Summary

...

## Changelog

```
```

## Relevant Issues

Fixes #5726 

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/components/popups/StopVotingPopup.svelte
```diff
@@ -31,13 +32,17 @@
         <TextHint info text={localize('popups.stopVoting.hint')} />
     </div>
     <div class="flex w-full space-x-4 mt-6">
-        <Button outline classes="w-full" onClick={onCancelClick}>{localize('actions.cancel')}</Button>
+        <Button outline classes="w-full" disabled={isTransferring} onClick={onCancelClick}
```
maxwellmattryan, line 35: Same as above ^

packages/shared/components/popups/ManageVotingPowerPopup.svelte
```diff
@@ -97,15 +100,10 @@
         <TextHint info text={localize('popups.manageVotingPower.hint')} />
     </div>
     <div class="flex flex-row flex-nowrap w-full space-x-4">
-        <Button outline classes="w-full" onClick={onCancelClick}>
+        <Button outline disabled={isTransferring} classes="w-full" onClick={onCancelClick}>
```
maxwellmattryan, line 103: Why would we want to disable the cancel button that closes the popup if the user can click the "X" anyway?

packages/shared/components/popups/VoteForProposalPopup.svelte
```diff
@@ -62,7 +63,9 @@
         {/if}
     </div>
     <popup-buttons class="flex flex-row flex-nowrap w-full space-x-4">
-        <Button classes="w-full" outline onClick={closePopup}>{localize('actions.cancel')}</Button>
+        <Button classes="w-full" disabled={isTransferring} outline onClick={closePopup}
```
maxwellmattryan, line 66: Same as above^

packages/shared/lib/contexts/governance/stores/latest-governance-transaction-ids.store.ts
```diff
@@ -0,0 +1,18 @@
+import { writable } from 'svelte/store'
+
+export const latestGovernanceTransactionIds = writable<{ [accountId: number]: string }>({})
```
maxwellmattryan, line 3: I think this name is confusing as it could mean the `latestGovernanceTransactionIds` that were confirmed. IMO `pendingGovernanceTransactionIds` would be more accurate since one it is confirmed then it is removed, meaning it's no longer pending.

packages/shared/lib/core/profile/actions/active-profile/logout.ts
```diff
@@ -1,6 +1,7 @@
 import { closePopup } from '@auxiliary/popup'
 import { resetSelectedAccount } from '@core/account'
-import { clearPollGovernanceDataInterval } from '@contexts/governance'
+import { clearPollGovernanceDataInterval } from '@contexts/governance/actions'
+import { resetLatestGovernanceTransactionIds } from '@contexts/governance/stores'
 import { isPollingLedgerDeviceStatus, stopPollingLedgerNanoStatus } from '@core/ledger'
 import { clearPollMarketPrices } from '@core/market/actions'
 import { clearPollNetworkInterval } from '@core/network'
```
maxwellmattryan, line 7: Might as well adjust the import specificity since we're here

maxwellmattryan (changes requested):
Looks good! Added a few questions and requested changes :pray:

maxwellmattryan:
Some feedback from testing:
- (Slightly unrelated but) I got a no unspent voting output found when trying to revote for proposals (after reducing voting power from 1000 to 500 SMR):
    ```
    No unspent voting output found
    ```
    Until I logged out and back in, my voting power went to 0 SMR even though it was reflected in my available balance (500 SMR). When I logged back in it displayed the correct voting power of 500 SMR.  
    Since I didn't revote, my expectation was that voting on a proposal would only vote for that single one, but it seems that all my previous votes were remembered. This is convenient for the user since they aren't required to manually re-vote for everything, but it's not the expectation.
- UX-wise, it would feel better if we also disabled the "Cancel" button in the `ProposalDetailsView` since we do it for the popups.

packages/shared/components/popups/ManageVotingPowerPopup.svelte
nicole-obrien: Because cancel to the user means that it will cancel an action, if you have already kicked off the action then the cancel button is not cancelling anything.

So its confusing to the user if it doesn't do anything. Why would we have a cancel button that only closes a popup when there is a close button.

packages/shared/components/popups/StopVotingPopup.svelte
nicole-obrien: Because cancel to the user means that it will cancel an action, if you have already kicked off the action then the cancel button is not cancelling anything.

So its confusing to the user if it doesn't do anything. Why would we have a cancel button that only closes a popup when there is a close button.

packages/shared/components/popups/VoteForProposalPopup.svelte
nicole-obrien: Because cancel to the user means that it will cancel an action, if you have already kicked off the action then the cancel button is not cancelling anything.

So its confusing to the user if it doesn't do anything. Why would we have a cancel button that only closes a popup when there is a close button.

packages/shared/lib/core/profile/actions/active-profile/logout.ts
nicole-obrien: huh?

nicole-obrien:
> * UX-wise, it would feel better if we also disabled the "Cancel" button in the `ProposalDetailsView` since we do it for the popups.

There is a task to remove this redundant button and replace it with the unvote button so this PR is fine for now

packages/shared/lib/contexts/governance/stores/latest-governance-transaction-ids.store.ts
nicole-obrien: Makes sense

maxwellmattryan (approved):
LGTM :fire:

PR #5773, opened 2023-02-06
enhancement: remove trailing spaces on url input

MarkNerdi:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


JannemanDev:
I would not strip *all* spaces in an url but only trailing and leading spaces. As a best practise (secutiry wise but also unwanted side effects) you should not change the url if that changes the location.

MarkNerdi:
> I would not strip _all_ spaces in an url but only trailing and leading spaces. As a best practise (secutiry wise but also unwanted side effects) you should not change the url if that changes the location.

it wouldnt really change the location, because a url with spaces is invalid and it would just correct it. 
But i agree that i'm gonna remove this, because its a bit shady if we manipulate the user input, whereas they should check and change it themselves

Tuditi (commented):
Maybe we should trim eventID as well?

Tuditi (approved):
LGTM! :fire:

PR #5883, opened 2023-02-15
refactor: TitleBar component

jeeanribeiro:
## Summary
Moves TitleBar component from shared to desktop and refactors its code

### Platforms
-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [x] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

## Checklist
-   [x] I have commented on my code, particularly in hard-to-understand areas


packages/shared/lib/auxiliary/icon/constants/icon-svg.ts
```diff
@@ -126,7 +132,7 @@ import {
 } from '../svgs'
 import { VERIFICATION_STATUS_OFFICIAL } from '../svgs/verification-status-official.svg'
 
-export const ICON_SVG = {
+export const ICON_SVG: { [key in Icon]: ISvg } = {
```
maxwellmattryan, line 135: I think we could rename to `ICON_SVG_MAP` to help be more descriptive and could also define a `IconSvgMap` type.

packages/shared/lib/auxiliary/icon/enums/icon.enum.ts
```diff
@@ -87,6 +91,7 @@ export enum Icon {
     Receive = 'receive',
     Refresh = 'refresh',
     Reset = 'reset',
+    RestoreSize = 'RestoreSize',
```
maxwellmattryan, line 94: Seems like `kebab-case` is the correct style to use for this enum definition?

maxwellmattryan (changes requested):
Code looks good, although the account switcher seems to have disappeared :thinking: 

Was also taking a look at the component itself and how it's used in `App.svelte` - it seems to be more than just a `TitleBar` since it includes the top nav plus the slot for the contents of the app. It would be nice to split this up into two components while we're here

maxwellmattryan (approved):
LGTM :fire:

Issue #5698, opened 2023-01-30
[Task]: Create reusable confirmation drawer


begonaalvarezd:
### Task description

Similar to what desktop uses with the confirmation popup, eg
```ts
openPopup({
    type: 'confirmation',
    props: {
        title: localize('actions.refreshTokenMetadata'),
        hint: localize('general.refreshTokenMetadataHint'),
        warning: true,
        confirmText: localize('actions.reset'),
        onConfirm: refreshTokenMetadata,
    },
})```
if would be nice having something similar in mobile to reuse code, which we can add eveywhere eg `DeleteAccountConfirmationView`, `NetworkConfigurationNodeConfirmationView`, ....

### Requirements

above mentioned

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #5886, opened 2023-02-15
refactor: mobile activity drawers

begonaalvarezd:
## Summary

This PR aims to kickoff the mobile drawers refactor, with the Activity Drawer. 
Drawers with routers are still missing, thought we could start with something easier.
The following logics involving drawers have been implemented:
- Select Activity
- Claim
- Reject
- Fast Claim
- Fast Reject
- Unlock Stronghold
- Confirm

Close https://github.com/iotaledger/firefly/issues/5698
Close https://github.com/iotaledger/firefly/issues/5879

## Changelog

```
- refactor: mobile activity drawers
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/mobile/App.svelte
```diff
@@ -14,7 +14,7 @@
         shouldBeDarkMode,
     } from '@core/app'
     import { onboardingProfile } from '@contexts/onboarding'
-    import { Route, ToastContainer } from './components'
+    import { Route, ToastContainer, DrawerManager } from './components'
```
maxwellmattryan, line 17: We can / should use the `@components` path alias. Might need to add a wildcard entry so that we can import from specific paths within `@components`.

packages/mobile/components/drawers/ConfirmDrawer.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    import { localize } from '@core/i18n'
+    import { Button, ButtonVariant, Text, TextHint } from 'shared/components'
```
maxwellmattryan, line 3: We should use `@ui` here

packages/mobile/components/drawers/DrawerManager.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    import { localize } from '@core/i18n'
+    import { Drawer } from '../../components'
```
maxwellmattryan, line 3: Feels strange to have an import like this from within the components folder :thinking: it's a strange one because it should be an atom since it's a generic component, but it's only mobile-specific...

packages/mobile/components/drawers/DrawerManager.svelte
```diff
@@ -0,0 +1,32 @@
...
+    function getStaticTitle(drawerId: DrawerId): string {
+        switch (drawerId) {
+            case DrawerId.EnterPassword:
+                return localize('popups.password.title')
+            case DrawerId.SelectedActivity:
+                return localize('popups.transactionDetails.title')
+            default:
+                return undefined
+        }
+    }
```
maxwellmattryan, line 24: Hmm what if we put this in the `Drawer` itself and pass the ID. Seems to be consistent we should also create a new locale category for `drawers` similar to popup, but we just make sure to use the same property as the ID so we don't have to write this logic.

packages/mobile/components/drawers/DrawerManager.svelte
```diff
@@ -0,0 +1,32 @@
...
+</script>
+
+{#each $drawersStore as drawer}
+    {@const drawerId = drawer.id}
+    <Drawer title={getStaticTitle(drawerId)} {...drawer.props} onClose={() => closeDrawer(drawerId)}>
+        <svelte:component this={COMPONENTS[drawerId]} {...drawer.props} />
+    </Drawer>
+{/each}
```
maxwellmattryan, line 32: How does this handle stacked drawers?

packages/mobile/components/drawers/EnterPasswordDrawer.svelte
```diff
@@ -0,0 +1,55 @@
+<script lang="ts">
+    import { localize } from '@core/i18n'
+    import { unlockStronghold } from '@core/profile'
+    import { Button, HTMLButtonType, PasswordInput, Text, TextType } from 'shared/components'
```
maxwellmattryan, line 4: We can use `@ui` here

packages/mobile/components/index.js
```diff
@@ -24,3 +24,4 @@ export * from './filters'
 export * from './inputs'
 export * from './labels'
 export * from './molecules'
+export * from './drawers'
```
maxwellmattryan, line 27: Should be alphabetized if possible :pray:

packages/mobile/lib/auxiliary/drawer/actions/updateDrawerProps.action.ts
```diff
@@ -0,0 +1,15 @@
+import { DrawerId } from '../enums'
+import { drawersStore } from '../stores'
+import { DrawerProps } from '../types'
+
+export function updateDrawerProps(drawerId: DrawerId, props?: DrawerProps): void {
```
maxwellmattryan, line 5: This function should go inside the corresponding store file since it deals directly with data management. We can also call it `updateDrawerState` to be consistent with the other CRUD functions.

packages/mobile/lib/auxiliary/drawer/actions/closeDrawer.action.ts
```diff
@@ -0,0 +1,6 @@
+import { DrawerId } from '../enums'
+import { drawersStore } from '../stores'
+
+export function closeDrawer(drawerId: DrawerId): void {
+    drawersStore.update(($drawersStore) => $drawersStore.filter((drawer) => drawer.id !== drawerId))
```
maxwellmattryan, line 5: Might be good to extract this line into a `removeDrawerState` function inside the store file, then call it here. It might prevent logic from being duplicated later if we need to remove a drawer state for some reason.

packages/mobile/lib/auxiliary/drawer/actions/openDrawer.action.ts
```diff
@@ -0,0 +1,8 @@
+import { DrawerId } from '../enums'
+import { drawersStore } from '../stores'
+import { DrawerProps } from '../types'
+
+export function openDrawer(drawerId: DrawerId, props?: DrawerProps): void {
```
maxwellmattryan, line 5: Same as the comment for `closeDrawer` - we can extract the logic here to a `addDrawerState` function that is called in `openDrawer`. Would be good in case we ever need to do extra logic besides actually updating the store itself.

packages/mobile/lib/auxiliary/drawer/enums/drawer-id.enum.ts
```diff
@@ -0,0 +1,14 @@
...
+    SelectedActivity = 'selected-activity',
+    Profile = 'profile',
+    Send = 'send',
+    Receive = 'receive',
+    SelectedToken = 'selected-token',
+    Legal = 'legal',
+    EnterPassword = 'enter-password',
+    AddReferences = 'add-references',
+    AddExpiration = 'add-expiration',
+    Confirm = 'confirm',
```
maxwellmattryan, line 13: Do these have to be `kebab-case`? The usual convention for enums is `PascalCase` for the key then `camelCase` for the value.

packages/mobile/lib/auxiliary/drawer/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './actions'
+export * from './enums'
+export * from './interfaces'
+export * from './stores'
+export * from './utils'
+export * from './types'
```
maxwellmattryan, line 6: `types` should be before `utils`

packages/mobile/lib/auxiliary/drawer/interfaces/drawer.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { DrawerId } from '../enums'
+import { DrawerProps } from '../types'
+
+export interface IDrawer {
```
maxwellmattryan, line 4: I think renaming this to `IDrawerState` might help it be more consistent with the store.

packages/mobile/lib/auxiliary/drawer/stores/drawers.stores.ts
```diff
@@ -0,0 +1,4 @@
+import { writable } from 'svelte/store'
+import { IDrawer } from '../interfaces'
+
+export const drawersStore = writable<IDrawer[]>([])
```
maxwellmattryan, line 4: I think `drawers` or `drawerStates` might be a better name.

packages/mobile/lib/contexts/dashboard/actions/handleClaimActivity.action.ts
```diff
@@ -0,0 +1,16 @@
+import { isStrongholdUnlocked } from '@core/profile-manager'
+import { Activity, claimActivity } from '@core/wallet'
+import { closeDrawer, DrawerId, openDrawer } from '../../../auxiliary/drawer'
+
+export async function handleClaimActivity(activity: Activity): Promise<void> {
```
maxwellmattryan, line 5: Wondering if this function belongs in a new `activity` context rather than `dashboard`, I think that might be a little too vague. Alternatively `wallet` would also make sense if you wanted to combine an `activity` and `token` context.

packages/mobile/lib/contexts/dashboard/actions/handleRejectActivity.action.ts
```diff
@@ -0,0 +1,19 @@
+import { localize } from '@core/i18n'
+import { rejectActivity } from '@core/wallet'
+import { closeDrawer, DrawerId, openDrawer } from '../../../auxiliary/drawer'
+
+export function handleRejectActivity(activityId: string): void {
```
maxwellmattryan, line 5: Same as above ^^^

packages/mobile/lib/auxiliary/drawer/actions/closeDrawer.action.ts
maxwellmattryan: Ah and we also don't need `.action` in the filename - for actions we use `camelCase` (reason is that's just what we started doing initially). If using `kebab-case` then we should be specifying the file type in the filename.

packages/mobile/views/dashboard/tabs/ActivityTab.svelte
```diff
@@ -1,20 +1,30 @@
...
-        $activityRouter?.next({ activity })
+        // TODO: when token drawer is refactor, implement this
+        // if (activity.type === ActivityType.Basic) {
+        //     const asset = getAssetById(activity.assetId)
+        //     // TODO: open token drawer to verify/skip token
+        // } else {
+        //     openDrawer(DrawerId.SelectedActivity, {
+        //         activity,
+        //     })
+        // }
```
maxwellmattryan, line 16: Hmm why would this logic be a part of the tab component?

maxwellmattryan (changes requested):
Looking good so far! Just have a few questions / comments :pray:

packages/mobile/components/drawers/DrawerManager.svelte
```diff
@@ -0,0 +1,32 @@
...
+    import { closeDrawer, DrawerId, drawersStore } from '../../lib/auxiliary/drawer'
+    import ConfirmDrawer from './ConfirmDrawer.svelte'
+    import EnterPasswordDrawer from './EnterPasswordDrawer.svelte'
+    import SelectedActivityDrawer from './SelectedActivityDrawer.svelte'
+
+    const COMPONENTS = {
+        [DrawerId.Confirm]: ConfirmDrawer,
+        [DrawerId.EnterPassword]: EnterPasswordDrawer,
+        [DrawerId.SelectedActivity]: SelectedActivityDrawer,
+    }
```
paul-boegelsack, line 13: Maybe we create a constant file for this, so we do not have to touch the manager every time we add a drawer?

packages/mobile/components/drawers/DrawerManager.svelte
paul-boegelsack: Maybe we create a constant file for this and add an object for each drawer: 

```
[Drawer.ID.EnterPasswordDrawer]: {
    component: EnterPasswordDrawer,
   title: localize('popups.password.title')
}
```

So we don't have to touch the manager every time we add a drawer

paul-boegelsack (changes requested):
I like this new solution! 🔥  Just have little things

packages/mobile/App.svelte
begonaalvarezd: I dont use mobile path aliases anywhere because they dont work out of the box because we use the desktop folder to run mobile, so adding new path aliases means tweaking the setup locally to patch desktop to use mobile path aliases

packages/mobile/components/drawers/DrawerManager.svelte
paul-boegelsack: Maybe we can create a drawer constant file where we connect the title and component with an id? So we do not have to change the manager every time we add a drawer.

packages/mobile/components/drawers/DrawerManager.svelte
begonaalvarezd: the drawer is very much a mobile only component, the import is terrible because we dont have path aliases working on mobile yet, i usually add a bit of redundancy in the import to find it easier in the future when we include the path aliases

packages/mobile/components/drawers/DrawerManager.svelte
begonaalvarezd: you can already test it, they get added consequently and because they are fixed and in the DOM they go in the order of the drawer store, they stack

packages/mobile/lib/auxiliary/drawer/stores/drawers.stores.ts
begonaalvarezd: I thought we started adding `Store` to every store to showcase is a store

packages/mobile/components/drawers/DrawerManager.svelte
begonaalvarezd: this reminds me, I'll remove the drawer we have in the shared folder

packages/mobile/components/drawers/DrawerManager.svelte
begonaalvarezd: I can move the static title to the drawer component, but we cant use the ID to get the static title (it can also be dynamic on multi-step drawers) because in mobile, the modals are also popups, so in the localization file not all strings are in popup, for example wallet switcher, create account, filters...

packages/mobile/lib/contexts/dashboard/actions/handleClaimActivity.action.ts
begonaalvarezd: moved to wallet, also opened an issue to refactor the dashboard context, it should be wallet

packages/mobile/views/dashboard/tabs/ActivityTab.svelte
begonaalvarezd: this logic in not in a component per se (though every .svelte files is a component ), this file is the full activity tab view, like a view parent that takes care of all the "routing" logic, in this case, the drawer logic. This logic is then passed to the `ActivityList` component, I dont see why is wrong having this logic here, what would you suggest instead?

begonaalvarezd: the logic is uncommented and implemented on its corresponding PR already

jeeanribeiro (changes requested):

packages/mobile/components/drawers/DrawerManager.svelte
```diff
@@ -0,0 +1,20 @@
...
+</script>
+
+{#each $drawers as drawer}
+    {@const drawerId = drawer.id}
```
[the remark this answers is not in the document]
begonaalvarezd: is just to avoid writing `drawer.id` N times

packages/mobile/components/drawers/ConfirmDrawer.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    import { localize } from '@core/i18n'
+    import { Button, ButtonVariant, Text, TextHint } from '@ui'
+
+    export let description: string = ''
+    export let hint: string = ''
+    export let info: boolean = false
+    export let success: boolean = false
+    export let warning: boolean = false
+    export let danger: boolean = false
```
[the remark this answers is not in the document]
begonaalvarezd: I would refactor this at another time

paul-boegelsack (approved):
lgtm 🚀

jeeanribeiro (approved):
LGTM, awesome work you have done 👏

PR #5889, opened 2023-02-15
refactor: Route component

jeeanribeiro:
## Summary
This PR removes the Route component

## Changelog

```
- Removes Route component
- Adapts code that was using the Route component
```

### Platforms
-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [x] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

## Checklist
-   [ ] I have commented on my code, particularly in hard-to-understand areas


packages/shared/components/Route.svelte
```diff
@@ -1,9 +1,11 @@
...
-    import { appRoute, AppRoute } from '@core/router'
+    import { AppRoute } from '@core/router'
+    import { Writable } from 'svelte/store'
 
     export let route: AppRoute
+    export let routeStore: Writable<AppRoute>
 </script>
 
-{#if $appRoute === route}
+{#if $routeStore === route}
```
begonaalvarezd, line 9: this component is so simple, that im thinking if we actually need it, we dont use it anywhere else because we manage the routing directly in the routing views, so Im thinking we might be able to get rid of this component altogether. We had it before because we used to have animations between routing transitions, but now i dont know if it adds value, wdyt?

jeeanribeiro: Removed the Route component

begonaalvarezd (approved):
🔥 🚒

PR #5916, opened 2023-02-17
Fix: glitch when sending output

Tuditi:
## Summary

Fixed the glitch when the dashboard router was reset from within the send confirmation popup. 

...

## Changelog

```
- Removed dashboardRouter.reset()
- Added check if NFT was removed and go back if so
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/shared/lib/core/wallet/actions/sendOutput.ts
+++ b/packages/shared/lib/core/wallet/actions/sendOutput.ts
@@ -1,9 +1,8 @@
 import { get } from 'svelte/store'
-import { NFT_OUTPUT_TYPE } from '@iota/types'
 import { selectedAccount, updateSelectedAccount } from '@core/account'
 import { removeNftFromSelectedAccount } from '@core/nfts/actions'
 
-import { DEFAULT_TRANSACTION_OPTIONS } from '../constants'
+import { DEFAULT_TRANSACTION_OPTIONS, OUTPUT_TYPE_NFT } from '../constants'
 import { resetNewTokenTransactionDetails } from '../stores'
 import { Output } from '../types'
 import { processAndAddToActivities } from '../utils'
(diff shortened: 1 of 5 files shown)


packages/shared/lib/core/wallet/actions/sendOutput.ts
```diff
@@ -11,6 +13,10 @@ export async function sendOutput(output: Output): Promise<void> {
         const account = get(selectedAccount)
         const transaction = await account.sendOutputs([output], DEFAULT_TRANSACTION_OPTIONS)
         // Reset transaction details state, since the transaction has been sent
+        if (output.type === OUTPUT_TYPE_NFT) {
+            updateNftInAllAccountNfts(account.index, output.nftId, { isSpendable: false })
```
MarkNerdi, line 17: this wont work when sending a newly minted output because `output.nftId` will be `0x000...`.
therefore we would need to get the nftId with the `getNftId` utils function. 

but because this one requires the output id and `output`  doesnt contain this, we should use the NFT output from the `transaction`, beacuse at this point, we know that `outputs.nftId` would contain the real value

Tuditi: I cannot reproduce this error. I think we're confusing it with handling transaction inclusions, since we are responsible for creating the outputs and `prepareOutput()` sets the `output.nftId` value.

MarkNerdi: confirmed :heavy_check_mark:

MarkNerdi (approved):
LGTM

PR #5991, opened 2023-02-28
feat: enable auto update

MarkNerdi:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/desktop/electron-builder-config.js
```diff
@@ -126,11 +138,7 @@ const prereleaseNsisOptions = {
...
     const icons = getIconPaths()
-    const publish = {
-        publishAutoUpdate: false,
-    }
-
-    return merge({}, prodConfig(), icons, { appId: APP_ID }, prereleaseNsisOptions, { publish })
+    return merge({}, prodConfig(), icons, { appId: APP_ID }, prereleaseNsisOptions)
 }
 
 const build = () => {
```
MarkNerdi, line 144: is there a reason why we use `testConfig` for `alpha` and `beta`s

packages/desktop/electron-builder-config.js
```diff
@@ -94,8 +106,8 @@ const prodConfig = () => ({
     publish: {
         provider: 'generic',
         url: 'https://dl.firefly.iota.org/',
```
begonaalvarezd, line 108: @rajivshah3 @MarkNerdi could we upload to a subfolder so we could use eg `https://dl.firefly.iota.org/shimmer/latest.yml` instead of  `https://dl.firefly.iota.org/shimmer.yml`, and upload the artifacts also to that subfolder, so the bucket is a bit more organized? Just a though here, not really important or a blocker for this PR to get merged 🙏🏼

MarkNerdi: image
i think it is possible if we change this location in the `build-desktop` file

nicole-obrien (changes requested):
Just waiting on a feature flag per os, so we can disable this on windows until there is a security fix or workaround for electron-updater on windows.

packages/desktop/electron/lib/menu.js
```diff
@@ -80,7 +80,7 @@ const buildTemplate = () => {
                 {
                     label: `${state.strings.checkForUpdates}...`,
                     click: () => getOrInitWindow('main').webContents.send('menu-check-for-update'),
-                    enabled: app.isPackaged ? state.enabled : false,
+                    enabled: app.isPackaged && features?.electron?.autoUpdates?.enabled ? state.enabled : false,
```
nicole-obrien, line 83: This popup is essentially a version popup so we still want to enable getting to it

packages/desktop/components/popups/CheckForUpdatesPopup.svelte
```diff
@@ -40,7 +46,7 @@
             keyText={localize('popups.appUpdate.stage')}
             valueText={localize(`popups.appUpdate.${get(appStage)}`)}
         />
-        {#if $appVersionDetails.upToDate}
+        {#if $appVersionDetails.upToDate || !hasAutoUpdate}
```
nicole-obrien, line 49: Perhaps if auto-update disabled then we show a hint to let the user know

packages/desktop/components/popups/CheckForUpdatesPopup.svelte
```diff
@@ -26,6 +31,7 @@
             await setAppVersionDetails()
             checkForAppUpdate()
         }
+        hasAutoUpdate = features.electron.autoUpdate[$platform]?.enabled
```
nicole-obrien, line 34: Should we not just wrap the checkForAppUpdate with this feature flag? we also do this in the App svelte file so we need to add the feature flag there

nicole-obrien (changes requested):

packages/desktop/components/popups/CheckForUpdatesPopup.svelte
nicole-obrien: Ahh okay that would make sense, in that case we should direct them to the firefly website

nicole-obrien (approved):
LGTM!

packages/desktop/electron-builder-config.js
nicole-obrien: Lets look at this when we have more access to s3

packages/desktop/electron-builder-config.js
nicole-obrien: Because we have different logos and icons i thinl

PR #6123, opened 2023-03-09
feat: adds update stronghold logic

### LINKED ISSUE #6049: [Task]: 7. Add logic to update stronghold view
### Task description

Now that we have the UI for updating a stronghold we need to add the logic to perform this action. Involving a call to the [wallet.rs](http://wallet.rs/) to to migrate the stronghold snapshots.

### Requirements

- Add logic to migrate stronghold with [wallet.rs](http://wallet.rs/)
- Add logic to update profile data to state the new version of stronghold for the profile

### Creation checklist

### LINKED ISSUE #6052: [Task]: 10. Add backup stronghold view to update stronghold router
### Task description

We need to add the optional backup stronghold view into the stronghold router, to encourage people to save a new stronghold

### Requirements

Add backup stronghold view into the update stronghold router

### Creation checklist

jeeanribeiro:
## Summary
closes #6049 

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [x] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

## Checklist


packages/desktop/views/login/views/update-stronghold/views/UpdateStrongholdView.svelte
```diff
@@ -15,7 +20,12 @@
 
     async function onSubmit(): Promise<void> {
         try {
-            await unlockStronghold(password)
+            await $profileManager.updateStronghold(password)
```
MarkNerdi, line 23: i think we still need to call `unlockStronghold`

maxwellmattryan:
It's working perfectly! There's just two weird things happening:
- The styling of the toast notification when changing the password seems to be `undefined`
- Getting a strange error when logging in / creating a profile: `Error: s is not a function`

packages/desktop/views/update-stronghold/views/UpdateStrongholdView.svelte
```diff
@@ -45,17 +60,22 @@
...
         <form on:submit|preventDefault={onSubmit} id="update-stronghold-form">
-            {#if !isRecovery}
-                <PasswordInput bind:value={password} bind:error={passwordError} autofocus showRevealToggle />
-            {/if}
+            <PasswordInput
+                bind:value={password}
+                bind:error={passwordError}
+                autofocus
+                showRevealToggle
+                classes={isPasswordInputVisible ? '' : 'invisible'}
```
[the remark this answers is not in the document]
maxwellmattryan: Rather than using the variable for a CSS class, just use a `{#if isPasswordInputVisble} ... {/if}` block

packages/shared/lib/core/profile-manager/actions/updateStronghold.ts
```diff
@@ -0,0 +1,32 @@
+import { get } from 'svelte/store'
+
+import { getStorageDirectoryOfProfile } from '@core/profile'
+import { activeProfileId, updateActiveProfile } from '@core/profile/stores'
+import { api, destroyProfileManager } from '@core/profile-manager'
+import { getSecretManagerPath } from '@core/profile-manager/utils'
```
maxwellmattryan, line 6: These should be relative imports

maxwellmattryan (changes requested):

packages/desktop/views/update-stronghold/views/UpdateStrongholdView.svelte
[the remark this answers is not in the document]
jeeanribeiro: I tried it before using the class, but the problem is that it was not displaying the error when what is inside the block was being rendered

maxwellmattryan:
Some feedback / questions for login flow:
- The "Update and continue" button needs a spinner (at least it's missing during login flow but recovery flow has one)
- The "Skip and keep old password" might be better off with a spinner although it is a little shorter than the "Update and continue" one
- The "Skip and keep old password" button should be disabled when changing the password and vice versa 
- Why is the "Skip wallet backup" button disabled when the password is changed - the user can technically continue without it, so is this just to force them to re-export? 
- Would look better to remove the backticks from any errors that show up (specifically looking at the `\`Stronghold migration error: Failed to decrypt snapshot: incorrect password or corrupt data\`` error)

Some feedback / questions for recovery flow:
- The UX feels strange that I can enter the wrong password and not know that until I click "Update and continue" in the update Stronghold view. It would feel better if I could immediately know that I have the wrong password when I input it the first time (but now I understand why you added that change to `isPasswordInputVisible`)
- I got this error at the end of onboarding recovering from a newly created V3 Stronghold:
    ```
    `stronghold client error: Error loading client data. No data present`
    ```

packages/shared/lib/core/profile-manager/actions/updateStronghold.ts
```diff
@@ -0,0 +1,33 @@
+import { get } from 'svelte/store'
+
+import { getStorageDirectoryOfProfile } from '@core/profile'
+import { activeProfileId, updateActiveProfile } from '@core/profile/stores'
+import { destroyProfileManager } from '@core/profile-manager/actions'
+import { api } from '@core/profile-manager/api'
+import { getSecretManagerPath } from '@core/profile-manager/utils'
```
maxwellmattryan, line 7: These should be relative import paths; we don't want to use the path alias if we're in the same module.

maxwellmattryan (changes requested):
One requested change around import paths + UX feedback

packages/shared/lib/core/profile-manager/actions/updateStronghold.ts
jeeanribeiro: Oh sorry, I've understood it wrongly, fixing it right now

jeeanribeiro:
> Some feedback / questions for login flow:
> 
> * The "Update and continue" button needs a spinner (at least it's missing during login flow but recovery flow has one)

I agree, it has no differentiation of isBusy state between the two, might be a bug of style updating
> * The "Skip and keep old password" might be better off with a spinner although it is a little shorter than the "Update and continue" one

Will add the isBusy to this button as well 👍 
> * The "Skip and keep old password" button should be disabled when changing the password and vice versa

Will add the disabled state when the password is being changed 👍 
> * Why is the "Skip wallet backup" button disabled when the password is changed - the user can technically continue without it, so is this just to force them to re-export?

https://github.com/iotaledger/firefly/issues/6058
> * Would look better to remove the backticks from any errors that show up (specifically looking at the `\`Stronghold migration error: Failed to decrypt snapshot: incorrect password or corrupt data`` error)

Will add this change 👍 
> Some feedback / questions for recovery flow:
> 
> * The UX feels strange that I can enter the wrong password and not know that until I click "Update and continue" in the update Stronghold view. It would feel better if I could immediately know that I have the wrong password when I input it the first time (but now I understand why you added that change to `isPasswordInputVisible`)

That is correct, I will try to add this change 👍 
> * I got this error at the end of onboarding recovering from a newly created V3 Stronghold using the wrong password (no error if using the correct password):
>   ```
>   `stronghold client error: Error loading client data. No data present`
>   ```
>   There's no way to go back from this point in the flow, so we must tell the user that their password is wrong when they enter it back in "Import Stronghold" view

The fix above might fix this flow 👍

maxwellmattryan:
Last pieces of feedback:

- When changing passwords, a spinner for both buttons shows up where as it should only be for the "Change password" button

PR #6317, opened 2023-03-30
fix: disable vote button after voting

### LINKED ISSUE #6184: [Task]: Vote button is active right after voting
### Task description

After setting a vote on a `counting` proposal, after the vote has been confirmed, the `Vote` button should be disabled because the already set vote is selected.

[Screencast from 15-03-2023 11:12:11.webm](https://user-images.githubusercontent.com/105642810/225282431-ddeeb8e3-b78f-4ae5-ba97-ca5379c80be7.webm)

### Requirements

- [x]  Update vote button state on voting

### Creation checklist

jeeanribeiro:
## Summary
closes #6184 

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [x] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

## Checklist

-   [ ] I have commented on my code, particularly in hard-to-understand areas


maxwellmattryan (commented):
Perhaps this is just on my machine with my network connection, but there's a small window of time from when the spinner stops after voting and the button is disabled (< 1s), then it becomes disabled and the answer starts pulsing. This allows me to still vote if I am fast enough to click the "Vote" button

Tuditi (changes requested):
If I click on a Proposal that I voted on the following happens:

1) While the proposal status/answer is loading the submit vote button is active
2) Once everything is loaded the vote button becomes deactivated.

I think during 1) the button should be inactive, so we should add a condition that the button is only active if the proposal has been loaded.

Tuditi (changes requested):
I added a suggestion

packages/desktop/views/dashboard/governance/views/ProposalDetailsView.svelte
```diff
@@ -71,13 +72,25 @@
...
+    $: $selectedParticipationEventStatus, (textHintString = getTextHintString())
+
     $: isVotingDisabled =
+        !isLoaded ||
         !isProposalVotable($selectedProposal?.status) ||
-        !hasChangedAnswers(selectedAnswerValues) ||
-        hasSelectedNoAnswers(selectedAnswerValues)
+        hasSelectedNoAnswers(selectedAnswerValues) ||
+        isUpdatingVotedAnswerValues ||
+        isSelectedEqualVotedAnswers
```
maxwellmattryan, line 82: This name is just slightly unclear, maybe something like `areSelectedAndVotedAnswersEqual` or `hasSelectedNoNewAnswers` - I might be being a bit pedantic here :sweat_smile:

maxwellmattryan (commented):
Left a suggestion

maxwellmattryan:
One last piece of feedback before I think it's ready to be merged:

When I click the "Stop voting" button, the spinner shows for both this and the "Vote" button instead of just showing for the "Stop voting" button while keeping the "Vote" button disabled. There was also 1-2 seconds after stopping a vote where the "Stop voting" button was active / "Vote" was disabled then it switched.

packages/desktop/views/dashboard/governance/views/ProposalDetailsView.svelte
```diff
@@ -50,6 +50,8 @@
     let statusLoaded: boolean = false
     let overviewLoaded: boolean = false
     let openedQuestionIndex: number = -1
+    let isUpdatingVotedAnswerValues: boolean = false
+    let lastAction: 'vote' | 'stopVote'
```
maxwellmattryan, line 54: We have a [`GovernanceAction`](https://github.com/iotaledger/firefly/blob/develop/packages/shared/lib/core/wallet/enums/governance-action.enum.ts) enum we should use here.

maxwellmattryan (changes requested):
One requested change :pray:

maxwellmattryan:
Some minor feedback:
- When I voted, I had to re-navigate back to the proposal details view for the "Stop voting" button to become enabled. Once the tx confirmed and the pulsing icons showed up, the "Stop voting" button was disabled along with the "Vote" button.
- I stopped voting then voted again for a proposal (without changing views), but the pulsing icons didn't show up until I re-navigated back to that view.

maxwellmattryan (approved):
Working like a charm :fire:

PR #6350, opened 2023-04-03
enhancement: remove or download NFT media if MaxMediaSize setting is changed

jeeanribeiro:
## Summary
Adds delete or download nft files after max media size setting is changed

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [x] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

- Change the max media size settings and check if nfts are being downloaded and deleted

## Checklist


packages/desktop/electron/main.js
```diff
@@ -408,6 +408,16 @@ ipcMain.handle('copy-file', (_e, sourceFilePath, destinationFilePath) => {
...
 })
 
+ipcMain.handle('delete-nft', (_e, { filePath, accountIndex, nftId }) => {
+    const userPath = app.getPath('userData')
+    const directory = app.isPackaged ? userPath : __dirname
+    const src = path.resolve(`${directory}/__storage__/${filePath}`)
+
+    fs.rmdir(src, { recursive: true, force: true }, () => {
+        windows.main.webContents.send('nft-delete-done', { accountIndex, nftId })
+    })
```
MarkNerdi, line 418: you can use `fs.rmdirSync` here. With this, we can just await this call and dont need to do anything regarding events

MarkNerdi (changes requested):
We can rewrite the delete logic in such a way such that we dont need to emit events, but just await it.

With this, we dont need a deletion queue. This will simplify all of it

PR #6447, opened 2023-04-11
feature: add sidedrawer component

MarkNerdi996:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/desktop/features/network.features.ts
+++ b/packages/desktop/features/network.features.ts
@@ -6,7 +6,7 @@ const networkFeatures: INetworkFeatures = {
         enabled: true,
     },
     config: {
-        enabled: true,
+        enabled: false,
     },
 }
 


packages/desktop/views/dashboard/Dashboard.svelte
```diff
@@ -11,7 +11,7 @@
         saveActiveProfile,
     } from '@core/profile'
     import { appRouter, dashboardRoute } from '@core/router'
-    import { Idle } from 'shared/components'
+    import { Idle, NetworkSideDrawer } from 'shared/components'
```
maxwellmattryan, line 14: We can use the `@ui` path alias here :+1:

packages/shared/components/organisms/NetworkSideDrawer.svelte
```diff
@@ -0,0 +1,23 @@
...
+    import { isNetworkSideDrawerOpen } from '@core/network/stores'
+    import { SideDrawer, Text, TextType, Icon } from 'shared/components/'
+    import { Icon as IconEnum } from '@lib/auxiliary/icon'
+
+    function closeDrawer(): void {
+        $isNetworkSideDrawerOpen = false
+    }
+</script>
+
+{#if $isNetworkSideDrawerOpen}
```
maxwellmattryan, line 12: We'll likely need to come up with a more dynamic and robust solution, most likely something similar to what mobile is doing with its own drawers (reason being that we will need to have routers for the different flows).

For this PR, I think adding a `sideDrawerState` store (corresponding to `ISideDrawerState`) with just one attribute, `direction`, would be good (should also be encoded into a `SideDrawerDirection` enum). If the state is not `null` or `undefined` then we render the side drawer. I will think about this before we start adding all of the other views and how we can route them together :+1:

packages/shared/components/organisms/NetworkSideDrawer.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+    import { localize } from '@core/i18n'
+    import { isNetworkSideDrawerOpen } from '@core/network/stores'
+    import { SideDrawer, Text, TextType, Icon } from 'shared/components/'
```
maxwellmattryan, line 4: We should be able to use the `@ui` path alias here, although because may not work because this file is also within the `shared/components` folder.

packages/shared/components/atoms/SideDrawer.svelte
```diff
@@ -0,0 +1,53 @@
...
+    </panel>
+</drawer>
+
+<style type="text/scss">
+    panel {
+        @apply fixed w-full max-h-full;
+        @apply flex flex-col flex-auto overflow-hidden;
+        @apply pt-7 px-5;
+        @apply bg-white;
+        width: 420px;
```
maxwellmattryan, line 45: Nice :sunglasses: although not sure exactly how many pixels the width of the drawer should be... will find out tomorrow :+1:

packages/shared/components/atoms/SideDrawer.svelte
```diff
@@ -0,0 +1,53 @@
+<script lang="ts">
+    import { fade, fly } from 'svelte/transition'
+    import { appSettings } from '@core/app'
+
+    export let onClose: () => unknown = () => {}
```
maxwellmattryan, line 5: Might be good to create a type for these types of functions, e.g. `type EventFn = (..._args: unknown[]) => void` since we use them a lot. I think in most if not all cases they always are `void`.

packages/shared/components/atoms/SideDrawer.svelte
maxwellmattryan, line 1: If you create a `ISideDrawerState`, we can just pass that to this `SideDrawer` component itself, and it knows which `direction`, if it should `preventClose`, etc.

packages/shared/components/atoms/SideDrawer.svelte
```diff
@@ -0,0 +1,53 @@
...
+    >
+        <slot />
+    </panel>
+</drawer>
+
+<style type="text/scss">
+    panel {
+        @apply fixed w-full max-h-full;
+        @apply flex flex-col flex-auto overflow-hidden;
+        @apply pt-7 px-5;
```
maxwellmattryan, line 43: With this padding the title text is slightly closer to the left of the side drawer, making it look just slightly uneven. The padding from all sides (possibly not the bottom) should be the same. 

Some side drawers will aslo have a back button to the left of the title text, so it would be good to see what that looks like here (sorry if that wasn't a task requirement).

maxwellmattryan (changes requested):
Have some changes requested :pray:

packages/shared/components/atoms/SideDrawer.svelte
MarkNerdi: :smoking:  i just used the width specified in figma

packages/desktop/lib/side-drawer/index.ts
maxwellmattryan, line 1: We should follow mobile here and create an `auxiliary` folder with a `drawer` module

packages/desktop/components/NetworkConfigRouter.svelte
maxwellmattryan, line 1: This might be best to put in a hierarchy of folders, i.e. `desktop > components > drawers > network-config > ...` then we can have a `views` folder within that (or call it something else since it's not quite a "view") with the `NetworkConfigRouter.svelte` in the root of that.

maxwellmattryan, line 1: This might be best to put in a hierarchy of folders, i.e. `desktop > components > drawers > network-config > ...` then we can have a `views` folder within that (or call it something else since it's not quite a "view") with the `NetworkConfigRouter.svelte` in the root of that.

packages/desktop/lib/side-drawer/interfaces/side-drawer-state.interface.ts
```diff
@@ -0,0 +1,10 @@
+import { SideDrawerDirection, SideDrawerType } from '../enums'
+
+export interface ISideDrawerState {
+    active: boolean
+    type: SideDrawerType
```
maxwellmattryan, line 5: Might be best to change to `route` if we change to `SideDrawerRoute`

packages/desktop/views/dashboard/Dashboard.svelte
```diff
@@ -110,6 +112,7 @@
         <div class="flex flex-col h-full dashboard-w">
             <svelte:component this={tabs[$dashboardRoute]} on:next={$appRouter.next} />
         </div>
+        <SideDrawer />
```
maxwellmattryan, line 115: Thinking it may be more "correct" to put the conditional rendering block here rather than in the component itself, i.e.
```svelte
{#if $sideDrawerState?.active}
    <SideDrawer />
{/if}
```

packages/desktop/webpack.config.ts
```diff
@@ -46,6 +46,7 @@ const resolve = {
         '@core': path.resolve(__dirname, '../shared/lib/core'),
         '@features': path.resolve(__dirname, './features'),
         '@lib': path.resolve(__dirname, '../shared/lib'),
+        '@desktop': path.resolve(__dirname, './lib'),
```
maxwellmattryan, line 49: Nvm :sweat_smile:

maxwellmattryan (changes requested):
Have some requested changes :pray:

maxwellmattryan (approved):
LGTM :fire:

maxwellmattryan (approved):
LGTM :fire: no. 2

PR #6634, opened 2023-04-21
feat: add ISCP/EVM network support

maxwellmattryan:
## Summary

This PR adds the MVP functionality for connecting to an ISCP/EVM node.

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

1. Enable `config` network feature flag
2. Uncomment the code within `onSubmitClick` in `AddChainSideDrawer.svelte`
3. Build and start the dev instance `yarn build && yarn start`
4. Navigate to the "Add chain" side drawer

## Checklist


packages/shared/lib/core/network/classes/iscp-chain.class.ts
```diff
@@ -1,13 +1,15 @@
...
+import { IBlock, IChain, IChainStatus, IIscpChainConfiguration, IIscpChainMetadata } from '../interfaces'
+import { ChainConfiguration, ChainMetadata, Web3Provider } from '../types'
 
 export class IscpChain implements IChain {
     private readonly _provider: Web3Provider
-    private readonly _metadata: IIscpChainMetadata
+    private readonly _configuration: IIscpChainConfiguration
 
-    constructor(payload: IIscpChainMetadata) {
+    private _metadata: IIscpChainMetadata
```
MarkNerdi, line 10: is this still needed?

packages/shared/lib/core/network/classes/iscp-chain.class.ts
```diff
@@ -31,16 +33,38 @@ export class IscpChain implements IChain {
...
+        return this._configuration
+    }
+
+    getMetadata(): Promise<ChainMetadata> {
+        if (this._metadata) {
+            return this._metadata
+        } else {
+            this._metadata = <IIscpChainMetadata>{} // await this.fetchChainMetadata()
+            return this._metadata
+        }
```
MarkNerdi: not sure if this is still needed?

MarkNerdi (changes requested):
Just heads up on some unused code. 
If it is needed for the next steps, feel free to ignore this comment

packages/desktop/features/network.features.ts
```diff
@@ -6,9 +6,9 @@ const networkFeatures: INetworkFeatures = {
         enabled: true,
     },
     config: {
-        enabled: false,
+        enabled: true,
         addChain: {
-            enabled: false,
+            enabled: true,
```
MarkNerdi, line 11: revert

packages/shared/lib/core/network/classes/stardust-network.class.ts
```diff
@@ -0,0 +1,89 @@
...
+        return get(networkStatus) ?? {}
+    }
+
+    getChain(chainId: number): IChain {
+        return this._chains.find((chain) => chain.getConfiguration().chainId === chainId)
+    }
+
+    getChains(): IChain[] {
+        return this._chains
+    }
```
MarkNerdi, line 50: we can transform these getters into "real" getters in javascript style (ref: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get?retiredLocale=de)

```suggestion
    get metadata(): NetworkMetadata {
        return this._metadata
    }

    get status(): INetworkStatus {
        return get(networkStatus) ?? {}
    }

    get chain(chainId: number): IChain {
        return this._chains.find((chain) => chain.getConfiguration().chainId === chainId)
    }

    get chains(): IChain[] {
        return this._chains
    }
```

packages/shared/lib/core/network/classes/iscp-chain.class.ts
maxwellmattryan: My thinking was that we should store it but lazy load then cache it, so the first time that the metadata needs to be retrieved it fetches from the node then it returns the cache on subsequent method calls.

packages/shared/lib/core/network/classes/stardust-network.class.ts
maxwellmattryan: Yes that makes sense, although probably only for the non-async variables. Did some looking around it seems async getters are possible, but I think that'd mean we have to use it like `const metadata = await chain.metadata`, which is sorta strange for a property.

maxwellmattryan: I think it may look strange to have both usages for getting data alongside eachother (some require it because it takes a parameter). I'd prefer just to keep it consistent, but we can change it later if we need to make use of a getter method.

packages/shared/lib/core/network/classes/iscp-chain.class.ts
maxwellmattryan: I think we should keep for the ISCP chain metadata we can fetch later

MarkNerdi (approved):
LGTM :+1:

Issue #6688, opened 2023-04-29
[Task]: Replace `NETWORK_ADDRESS` constant with `DEFAULT_CHAIN_CONFIGURATIONS` for L1 to L2 transfers on ShimmerEVM


maxwellmattryan:
### Task description

This is an old constant that has been updated with the recent changes to support connecting to L2 chains.

### Requirements

_TBD_

### Creation checklist

- [x] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #6707, opened 2023-05-02
chore: merge network address into default chain configuration

Tuditi:
## Summary

closes: #6688 
...

## Changelog

```
- Removes NETWORK_ADDRESS constant
- Add isIscpChain & isEvmChain type guards
- Replace magic strings with DestinationNetwork enums
- Consistently showing DestinationNetwork in activities
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/network/classes/stardust-network.class.ts
```diff
@@ -13,14 +13,14 @@ import { IscpChain } from './iscp-chain.class'
...
-    private readonly _chains: IChain[]
+    private readonly _chains: (IChain | undefined)[]
 
     constructor(metadata: NetworkMetadata, chainConfigurations: ChainConfiguration[]) {
         this._metadata = metadata
         this._chains = this.constructChains(chainConfigurations ?? [])
     }
 
-    private constructChains(chainConfigurations: ChainConfiguration[]): IChain[] {
+    private constructChains(chainConfigurations: ChainConfiguration[]): (IChain | undefined)[] {
```
MarkNerdi, line 23: i think we shouldnt allow `undefined` to be in the chains. After the `.map` we should filter for all non-undefined elements

packages/shared/lib/core/network/constants/default-chain-configurations.constant.ts
```diff
@@ -1,17 +1,18 @@
+import { DestinationNetwork } from '@core/layer-2'
 import { ChainType, NetworkId } from '../enums'
 import { ChainConfiguration } from '../types'
 
 export const DEFAULT_CHAIN_CONFIGURATIONS: Readonly<{ [id in NetworkId]?: ChainConfiguration }> = {
     [NetworkId.Shimmer]: {
         type: ChainType.Iscp,
-        name: 'ShimmerEVM',
+        name: DestinationNetwork.Shimmer,
```
MarkNerdi, line 8: shouldnt it be `DestinationNetwork.ShimmerEvm`

packages/shared/lib/core/network/interfaces/network.interface.ts
```diff
@@ -6,8 +6,8 @@ export interface INetwork {
     getMetadata(): NetworkMetadata
     getStatus(): INetworkStatus
 
-    getChain(chainId: number): IChain
-    getChains(): IChain[]
+    getChain(chainId: number): IChain | undefined
+    getChains(): (IChain | undefined)[]
```
MarkNerdi, line 10: remove `unefined` here if you agree on the above comment

packages/shared/lib/core/network/utils/isEvmChain.ts
```diff
@@ -0,0 +1,6 @@
+import { IEvmChainConfiguration } from '../interfaces'
+import { ChainConfiguration } from '../types'
+
+export function isEvmChain(configuration: ChainConfiguration): configuration is IEvmChainConfiguration {
```
MarkNerdi, line 4: what exactly does `configuration is IEvmChainConfiguration` do?

packages/shared/components/inputs/NetworkInput.svelte
```diff
@@ -1,48 +1,49 @@
...
     import { Modal, SelectorInput, IOption } from 'shared/components'
     import { activeProfile, getNetworkHrp } from '@core/profile'
-    import { NETWORK_ADDRESS, DestinationNetwork } from '@core/layer-2'
     import { validateBech32Address } from '@core/utils'
+    import { isIscpChain } from '@core/network'
+    import type { ChainConfiguration } from '@core/network'
 
-    const readonlyAttribute = $activeProfile?.isDeveloperProfile ? {} : { readonly: false }
-    const networkAddresses = NETWORK_ADDRESS[$activeProfile?.network?.id]
+    export let layer2ChainAddress: string | undefined = undefined
```
MarkNerdi, line 8: is there a reason that you renamed this from  `networkAddress`? this var can both include a L1 and L2, no?

packages/shared/components/inputs/NetworkInput.svelte
```diff
@@ -1,48 +1,49 @@
...
-    $: networkAddress = selected?.value
+    function getNetworkOptions(showLayer2: boolean): IOption[] {
+        let layer2Networks: IOption[] = []
+        if (showLayer2) {
+            layer2Networks =
+                $activeProfile.network?.chainConfigurations?.map((chain) => ({
+                    key: chain.name,
+                    value: getNetworkValue(chain),
+                })) ?? []
+        }
```
MarkNerdi, line 35: should we show chains in the dropdown if the value is `undefined` ?

packages/shared/lib/core/network/utils/isEvmChain.ts
Tuditi: It's a type guard. It ensures that the type of `configuration` is narrowed down to `IEvmChainConfiguration`

packages/shared/lib/core/network/interfaces/network.interface.ts
Tuditi: I agree, but for l.9 it's a bit difficult, since `find` can return `undefined`. Alternatively, we can throw an error if the chain isn't found, but I'm not sure if that should happen within the class itself.

packages/shared/lib/core/network/constants/default-chain-configurations.constant.ts
Tuditi: Good catch!

packages/shared/components/inputs/NetworkInput.svelte
Tuditi: I renamed it because we could get the layer2ChainAddress from the `SendFormPopup`, but you're right, inside `NetworkInput` it should be called differently.

Tuditi: Not really, there is no L1 networkAddress. The network address (~alias address) only exists for ISCP Chains, so I'll change it to iscpChainAddress instead. If it's undefined then we have a L1 transaction. We need to figure out a way on how to handle this for EVM -> EVM transactions. But that's part of bigger refactor.

packages/shared/components/inputs/NetworkInput.svelte
Tuditi: Yeas, because the value defines the alias address, but that's only available for ISCP chains.

MarkNerdi (approved):
LGTM :fire:

packages/shared/components/inputs/NetworkInput.svelte
Tuditi: Not at the moment. Once we have EVM Chains, we need a better way of selecting the chain, maybe by introducing the `id` for the different options.

PR #6709, opened 2023-05-02
refactor: send flow

MarkNerdi:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/desktop/components/popups/SendFlowPopup.svelte
maxwellmattryan, line 1: This should be named `SendPopupRouter` to be consistent with the other router components we have

maxwellmattryan: From a DX point of view, having either `Send` or `SendRouter` as the `PopupId` makes sense too

packages/desktop/components/SendButton.svelte
```diff
@@ -4,13 +4,16 @@
     import { resetNewTokenTransactionDetails } from '@core/wallet'
     import { openPopup, PopupId } from '@auxiliary/popup'
     import { OnboardingButton } from '@ui'
+    import { SendFlowRouter, sendFlowRouter } from '@views/dashboard/send-flow'
```
maxwellmattryan, line 7: I'd argue that these files aren't views themselves - they're popups so they should exist alongside the popup components. I'm inclined to say the router code should exist within the `desktop/routers` module but it makes sense alongside the popup components as well.

packages/desktop/views/dashboard/index.js
```diff
@@ -3,5 +3,6 @@ export * from './developer'
 export * from './settings'
 export * from './wallet'
 export * from './governance'
+export * from './send-flow'
```
maxwellmattryan, line 6: This isn't a collection of views like the others here are, this should be in a `components/popups/send` directory IMO

packages/desktop/views/dashboard/send-flow/send-flow-route.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum SendFlowRoute {
+    SelectToken = 'selectToken',
+    InputTokenAmount = 'inputTokenAmount',
+    InputDestination = 'inputDestination',
+    TransactionSummary = 'transactionSummary',
```
maxwellmattryan, line 5: Admittedly the technical spec. isn't where it should be, but these names should be consistent with what is outlined by the tasks, i.e. `SelectAsset`, `SelectNetworkAndRecipient`, `SelectAmount` (alt. `SetAmount`), `SendDetails`. I do think `TransactionSummary` is nice though.

Regardless, whatever we go with here must be reflected in the technical spec and the tasks that we create that use component names.

packages/desktop/views/dashboard/send-flow/send-flow-router.ts
maxwellmattryan, line 1: Filename should be `send.router.ts` (given we remove "flow" from the name)

packages/desktop/views/dashboard/send-flow/views/InputDestinationView.svelte
maxwellmattryan, line 1: See comment about route names

packages/shared/lib/auxiliary/popup/enums/popup-id.enum.ts
```diff
@@ -33,6 +33,7 @@ export enum PopupId {
     RemoveNode = 'removeNode',
     RemoveProposal = 'removeProposal',
     Revote = 'revote',
+    SendFlow = 'sendFlow',
```
maxwellmattryan, line 36: Not sure why but it feels strange to me that we would have "Flow" in the name... :thinking:

maxwellmattryan (changes requested):
Have some requested change :pray:

maxwellmattryan:
Have some things to fix:
- When clicking back in the `SelectAssetPopup`, it's seeing the router as `undefined` (closing the popup is fine):
    ```
    Cannot read properties of undefined (reading 'previous')
    TypeError: Cannot read properties of undefined (reading 'previous')
        at SendFlowRouter.previous (http://localhost:8080/build/index.js:53788:25)
        at onBackClick (http://localhost:8080/build/index.js:197235:19)
        at http://localhost:8080/build/index.js:58179:42
    ```
- When clicking back in the `SelectNetworkAndRecipientPopup` I get this error for the amount (closing the popup is fine):
    ```
    Error: Amount is negative
        at formatTokenAmountDefault (formatTokenAmountDefault.ts:12:1)
        at instance (AssetAmountInput.svelte:28:1)
        at init (index.mjs:2122:1)
        at new AssetAmountInput (AssetAmountInput.svelte:40:1)
        at createProxiedComponent (svelte-hooks.js:341:1)
        at new ProxyComponent (proxy.js:242:1)
        at new Proxy<AssetAmountInput> (proxy.js:349:1)
        at create_fragment (InputTokenAmountView.svelte:48:1)
        at init (index.mjs:2137:1)
        at new InputTokenAmountView (InputTokenAmountView.svelte:51:1)
    ```
- Clicking continue in the `SelectNetworkAndRecipientPopup` gets the same error as directly above plus a client error from `wallet.rs`:
    ```
    {type: 'client', error: '`invalid amount in API response: -1`'}
    ```

packages/desktop/components/popups/SendFlowPopup.svelte
MarkNerdi: IMO having `SendFlowPopup` as the parent component name is better, because it describes best what this popup is for.

`SendPopupRouter` includes our technical implementation in its name, but i think the parent component name should be agnostic to that. Because at its core its just a "popup for the send flow"

packages/desktop/components/SendButton.svelte
MarkNerdi: i agree that the router  code should be in `desktop/routers` and the rest in `popups`

packages/desktop/views/dashboard/send-flow/send-flow-router.ts
MarkNerdi: agree! we should change that as well in all `onboarding` routes

maxwellmattryan (changes requested):
Still getting the same errors in the above comment

nicole-obrien (approved):

PR #6772, opened 2023-05-10
test: string utils

MarkNerdi:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist

### DIFF
--- a/packages/shared/lib/core/utils/string.ts
+++ b/packages/shared/lib/core/utils/string.ts
@@ -63,7 +63,7 @@ export function getInitials(name: string | undefined, maxChars?: number): string
         .split(' ')
         .filter((n) => n)
         .map((n) => n.match(/./gu)) // match characters for emoji compatibility
-        .filter((n): n is RegExpMatchArray => !n)
+        .filter((n): n is RegExpMatchArray => n !== null)
         .map((n) => n[0])
 
     if (maxChars !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/shared/lib/core/utils/string.ts
```diff
@@ -81,7 +66,7 @@ export function getInitials(name: string | undefined, maxChars: number): string
         .filter((n) => n)
```
kraftjs, line 66: I'm getting an error that n could be null in line 67 when accessing n[0] so I did some googling and found some possible solutions
```suggestion
        .filter((n): n is RegExpMatchArray => n !== null)
```
or you could use a flatMap `.flatMap((n) => n ? [n[0]] : [])` to do the same thing without needing to make use of type guard

sources:
https://stackoverflow.com/a/51577579
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

kraftjs: Otherwise, LGTM

kraftjs (approved):
LGTM

kraftjs:
Edited `.filter((n): n is RegExpMatchArray => !n)` to `.filter((n): n is RegExpMatchArray => n !== null)` because it was causing the tests Test check to fail

Issue #6762, opened 2023-05-08
[Task]: Display L2 assets in asset list


maxwellmattryan:
### Task description

The user must be able to see their assets from the main wallet dashboard.

### Requirements

_TBD_

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #6801, opened 2023-05-15
feat: fetch wrapped native tokens from a static evm address

begonaalvarezd:
## Summary

Close https://github.com/iotaledger/firefly/issues/6762
CLose https://github.com/iotaledger/firefly/issues/6832

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/layer-2/actions/getLayer2NativeTokens.ts
```diff
@@ -0,0 +1,48 @@
...
+                    id: id,
+                }
+
+                nativeTokens.push(nativeToken)
+            }
+        } catch (e) {
+            console.error(e)
+        }
+    })
+    return nativeTokens
```
MarkNerdi, line 47: i think if you use an async function in a foreach like this, in the outer function flow, `chains?.forEach` doesnt await for all iterations to be finished.

i would suggest something similar to this:
```suggestion
    const nativeTokensPromises = chains.map(chain => {
        try {
            const provider = chain.getProvider()
            const contract = new provider.eth.Contract(ISC_SANDBOX_ABI, ISC_CONTRACT)
            const nativeTokenResult = await contract.methods
                .callView(accountsCoreContract, getBalanceFunc, parameters)
                .call()

            for (const item of nativeTokenResult.items) {
                const id = item.key
                const idBytes = Converter.hexToBytes(id)

                if (idBytes.length !== TOKEN_ID_BYTE_LENGTH) {
                    continue
                }

                const nativeToken = {
                    amount: BigInt(item.value),
                    id: id,
                }

                return Promise.resolve(nativeToken)
            }
        } catch (e) {
            return Promise.reject()
        }
    })
    
    const nativeTokens = await Promise.all(nativeTokensPromises)
    return nativeTokens
```

packages/shared/lib/core/layer-2/utils/getSmartContractHexName.ts
```diff
@@ -0,0 +1,19 @@
...
+    const SMART_CONTRACT_HEX_NAME_LENGTH = 4
+    const stringBytes = Converter.utf8ToBytes(name)
+    const hash = Blake2b.sum256(stringBytes)
+
+    for (let i = 0; i < hash.length; i += SMART_CONTRACT_HEX_NAME_LENGTH) {
+        const slice = hash.slice(i, i + SMART_CONTRACT_HEX_NAME_LENGTH)
+        const cursor = new SimpleBufferCursor(Buffer.from(slice))
+
+        return cursor.readUInt32LE()
+    }
```
MarkNerdi, line 16: we return from this loop everytime at the first iteration, is this correct?

begonaalvarezd: yes you are right, I updated the code 🙏🏼

packages/shared/lib/core/layer-2/types/abi.type.ts
```diff
@@ -0,0 +1 @@
+export type AbiType = 'function' | 'constructor' | 'event' | 'fallback'
```
maxwellmattryan, line 1: Perhaps we should use an `enum` if possible, imo they're slightly more friendly to use than string union types.

packages/shared/lib/core/layer-2/types/state-mutability.type.ts
```diff
@@ -0,0 +1 @@
+export type StateMutabilityType = 'pure' | 'view' | 'nonpayable' | 'payable'
```
maxwellmattryan, line 1: Same as above

packages/shared/lib/core/layer-2/constants/isc-contract.constant.ts
```diff
@@ -0,0 +1 @@
+export const ISC_CONTRACT = '0x1074000000000000000000000000000000000000'
```
maxwellmattryan, line 1: This is the magic contract right? Would be slightly less ambiguous to have `ISC_MAGIC_CONTRACT_ADDRESS = ...`

packages/shared/lib/core/layer-2/interfaces/layer2-native-token.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { IIrc30Metadata } from '@core/wallet'
+
+export interface ILayer2NativeToken {
```
maxwellmattryan, line 3: What exactly is this an interface for? Is this like wrapped L1 native tokens on L2?

maxwellmattryan (changes requested):
Have some minor requested changes :pray:

MarkNerdi (approved):
looks good from my side :fire:

packages/shared/lib/core/layer-2/interfaces/layer2-native-token.interface.ts
begonaalvarezd: I didnt want to overcomplicate the PR making the PR work with the existing `IAsset` because we barely have any information at this point about the asset, so i found it easier using this type until we can format the full information of the asset into something we already have

packages/shared/lib/core/layer-2/types/abi.type.ts
begonaalvarezd: updated ✅

packages/shared/lib/core/layer-2/types/state-mutability.type.ts
begonaalvarezd: updated ✅

packages/shared/lib/core/layer-2/constants/isc-contract.constant.ts
begonaalvarezd: updated ✅

packages/shared/lib/core/layer-2/interfaces/layer2-native-token.interface.ts
begonaalvarezd: I'll update these types in the next PR to be able to add the assets to the persisted profiles

packages/shared/lib/core/layer-2/actions/getLayer2NativeTokens.ts
```diff
@@ -0,0 +1,51 @@
...
+import { TOKEN_ID_BYTE_LENGTH } from '@core/token'
+import { Converter } from '@iota/util.js'
+import { get } from 'svelte/store'
+import { ISC_MAGIC_CONTRACT_ADDRESS } from '../constants'
+
+// TODO: remove this hardcoded address and use the account's EVM address instead
+const HARDCODED_EVM_ADDRESS = 'WRITE_YOUR_EVM_ADDRESS_HERE'
+
+// TODO: missing set native tokens, which converts the function to getAndSetLayer2NativeTokens
+export async function getLayer2NativeTokens(): Promise<{ amount: bigint; id: string }[]> {
```
maxwellmattryan: We should change the name to `getWrappedNativeTokens` then we can see if we can re-use some code for `getNativeL2Tokens` or even `getL2Tokens` which would be a combination of both of those. All of these would be per chain

packages/shared/lib/core/layer-2/actions/pollLayer2NativeAssets.ts
```diff
@@ -0,0 +1,15 @@
+import { LAYER2_NATIVE_ASSETS_POLL_INTERVAL } from '../constants'
+import { getLayer2NativeTokens } from './'
+
+let pollInterval: number
+
+export function pollLayer2NativeAssets(): void {
```
maxwellmattryan, line 6: We should make a `pollLayer2Tokens` that will be responsible for getting both wrapped native tokens and native (to) L2 tokens.

maxwellmattryan (changes requested):

maxwellmattryan (approved):
LGTM :fire:

Issue #6891, opened 2023-05-22
[Task]: Fix evm ledger generation flow


begonaalvarezd:
### Task description

Fix ledger evm generation & adapt to new account persisted data

### Requirements

TBD

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #6871, opened 2023-05-22
feat: cache evm addresses

MarkNerdi:
## Summary

> Please summarize your changes, describing **what** they are and **why** they were made.

⚠️ depends on https://github.com/iotaledger/firefly/pull/6880

## Changelog

```
- Please write a list of changes
```

### Platforms

> Please select any platforms where your changes have been tested.

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

> Please describe the specific instructions, configurations, and/or test cases necessary to **test and verify** that your changes work as intended.

...

## Checklist

> Please tick the following boxes that are relevant to your changes.

-   [ ] I have followed the contribution guidelines for this project
-   [ ] I have performed a self-review of my own code
-   [ ] I have commented my code, particularly in hard-to-understand areas
-   [ ] I have added or modified tests that prove my changes work as intended
-   [ ] I have verified that new and existing unit tests pass locally with my changes
-   [ ] I have verified that my latest changes pass CI workflows for testing and linting
-   [ ] I have made corresponding changes to the documentation

PR #6892, opened 2023-05-22
fix: ledger evm address generation

begonaalvarezd:
## Summary

Close https://github.com/iotaledger/firefly/issues/6891
⚠️ depends on https://github.com/iotaledger/firefly/pull/6871

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


maxwellmattryan (approved):
Code all looks good, although I'm getting stuck when I try to generate an address - it never leaves the loading state and there's no prompt on the device

maxwellmattryan (changes requested):
Meant to request changes :sweat_smile:

begonaalvarezd:
> Code all looks good, although I'm getting stuck when I try to generate an address - it never leaves the loading state and there's no prompt on the device

@maxwellmattryan for me the infinite loading only happens if I have the ledger connected with the ethereum app open and **the device is locked**, in this case the ledger status we get is wrong and it appears to be unlocked when that is not true. 
I could have a look at this in this PR, the original scope of the PR was to fix the existing views to accomodate the new evm addresses persisted structure.

Here is the ledger state firefly detects when the status is:
- ethereum app open
- device locked (request to enter pin)

mind `locked: false` ⬅️ 

image

begonaalvarezd:
> > Code all looks good, although I'm getting stuck when I try to generate an address - it never leaves the loading state and there's no prompt on the device
> 
> @maxwellmattryan for me the infinite loading only happens if I have the ledger connected with the ethereum app open and **the device is locked**, in this case the ledger status we get is wrong and it appears to be unlocked when that is not true. I could have a look at this in this PR, the original scope of the PR was to fix the existing views to accomodate the new evm addresses persisted structure.
> 
> Here is the ledger state firefly detects when the status is:
> 
> * ethereum app open
> * device locked (request to enter pin)
> 
> mind `locked: false` arrow_left
> 
> image

@maxwellmattryan It seems to be an issue thats not resolvable on our end, I see the wrong status comes from `manager.getLedgerNanoStatus()`

begonaalvarezd:
Additionally, I just tested the status we get for all the relevant apps with
- ledger locked & screen on
- ledger locked & screen off

and I get the following:

iota app:
image

shimmer app:
image

ethereum app:
image

so its an issue with the status we get from the ethereum app

packages/shared/lib/core/layer-2/actions/loadEvmAddressForSelectedAccount.ts
```diff
@@ -1,22 +1,15 @@
...
         }
         const { evmAddresses, index } = account
-        const chains = get(network)?.getChains() ?? []
-        const coinTypes = new Set(chains.map((chain) => chain.getConfiguration().coinType))
-
-        for (const coinType of coinTypes) {
-            if (!evmAddresses[coinType]) {
-                await Ledger.generateEvmAddress(coinType, index, false)
-            }
+        if (coinType !== undefined && !evmAddresses[coinType]) {
```
maxwellmattryan, line 11: We can skip this first condition since we're already testing for it

maxwellmattryan (dismissed):
One tiny (technically unnecessary) change :pray:

packages/desktop/components/drawers/network-config/drawers/ConfirmLedgerEvmAddressDrawer.svelte
```diff
@@ -1,16 +1,19 @@
...
     import { DrawerRoute, NetworkConfigRoute, networkConfigRouter } from '@desktop/routers'
+    import { Animation, Button, CopyableBox, FontWeight, Pane, Text, TextType } from '@ui'
+    import { onDestroy, onMount } from 'svelte'
     import DrawerTemplate from './DrawerTemplate.svelte'
-    import { Router } from '@core/router'
 
     export let drawerRouter: Router<DrawerRoute>
 
-    $: address = $selectedAccount?.evmAddress
+    let coinType
```
MarkNerdi, line 15:
```suggestion
    let coinType: number | undefined
```

packages/shared/lib/core/layer-2/actions/loadEvmAddressForSelectedAccount.ts
```diff
@@ -1,22 +1,15 @@
...
 import { Ledger } from '@core/ledger/classes'
-import { network } from '@core/network'
-import { get } from 'svelte/store'
 
-export async function loadEvmAddressForSelectedAccount(): Promise<void> {
+export async function loadEvmAddressForSelectedAccount(coinType: number): Promise<void> {
     try {
         const account = getSelectedAccount()
-        if (!account) {
+        if (!account || coinType === undefined) {
```
MarkNerdi, line 7: we specify the function param as `number` which doesnt include undefined. we can remove this check

packages/desktop/components/drawers/network-config/drawers/ConfirmLedgerEvmAddressDrawer.svelte
begonaalvarezd: updated ✅

packages/shared/lib/core/layer-2/actions/loadEvmAddressForSelectedAccount.ts
begonaalvarezd: updated ✅

MarkNerdi (approved):
LGTM

PR #6897, opened 2023-05-23
enhancement: improve token amount input

MarkNerdi:
## Summary

...

## Changelog

```
- reduce the fontsize if amount succeeds a certain length
- If a decimal point is the first thing inputted, a 0 is added before the decimal point
- block inputting a number which would be longer than available balance
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/components/inputs/TokenAmountInput.svelte
```diff
@@ -1,28 +1,74 @@
...
     export let inputtedAmount: string | undefined =
         rawAmount && asset?.metadata
             ? formatTokenAmountDefault(Number(rawAmount), asset.metadata, unit, false)
             : undefined
 
     let amountInputElement: HTMLInputElement
     let error: string
+    let inputLength: number = 0
+    let fontSize: number = 64
+    let maxLength: number = 0
```
Tuditi, line 23: In cases where we initialize a variable as a number, it's redundant to type them. Typing is only required for exported variables, so that the parent components also throw errors if the type is not correctly adhered to.

The elements on l.19 & l.20 are `undefined` at initialization, so maybe that type should be added as well?

packages/shared/components/inputs/TokenAmountInput.svelte
```diff
@@ -42,22 +88,23 @@
...
+                        {fontSize}
                         clearBackground
                         clearPadding
                         clearBorder
                     />
                 </amount-wrapper>
             </div>
 
-            <Text fontWeight={FontWeight.semibold} classes="py-4">
+            <Text fontWeight={FontWeight.semibold} classes={inputLength < 14 ? 'py-4' : 'py-2'}>
```
Tuditi, line 107: What's the reason fir this conditional?

Tuditi (changes requested):
Added some minor request for changes.

How do you think we should handle very large numbers? E.g. 1.11e40

packages/shared/components/inputs/TokenAmountInput.svelte
MarkNerdi: image
here you see that the input has padding on the top and bottom. This `py-X` is here to align the `SMR` with that padding

MarkNerdi:
Regarding **How do you think we should handle very large numbers? E.g. 1.11e40**:

i think for now we shouldnt support these numbers in the send form, at least for now

Tuditi (approved):
LGTM! :fire:

Issue #6817, opened 2023-05-15
[Task]: Add support for caching ERC20 tokens


maxwellmattryan:
### Task description

We'll need to have this data so we can display various information about it to the user.

### Requirements

- [ ] Change persisted asset data to handle layer 2 tokens
- [ ] Change `IPersistedAsset` to `IPersistedToken`
- [ ] Write migration function if necessary
- [ ] Remove IRC27 standard from `TokenStandard`

### Creation checklist

- [x] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #6932, opened 2023-06-01
feat: persist erc-20 tokens

begonaalvarezd:
## Summary

Close https://github.com/iotaledger/firefly/issues/6817

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
```diff
@@ -1,12 +1,19 @@
...
+import { IErc20Metadata, IIrc30Metadata, IPersistedAsset } from '../interfaces'
 import { AssetVerification } from '../types'
-import { getIrc30MetadataFromFoundryOutput } from '../utils/getIrc30MetadataFromFoundryOutput'
+import { getIrc30MetadataFromFoundryOutput } from '../utils'
 
-export async function requestPersistedAsset(tokenId: string): Promise<IPersistedAsset> {
-    const tokenMetadata = await getIrc30MetadataFromFoundryOutput(tokenId)
+export async function requestPersistedAsset(tokenId: string, chainId?: number): Promise<IPersistedAsset | undefined> {
+    let tokenMetadata: IIrc30Metadata | IErc20Metadata | undefined
+    if (chainId && isValidEthereumAddress(tokenId)) {
```
MarkNerdi, line 12: just to make sure: is this function really differentiating between erc20 ids and irc30 ids? both of them look like "0x...".
is the differentiating factor the checksum?

packages/shared/lib/core/utils/crypto/utils/isValidEthereumAddress.ts
maxwellmattryan, line 1: We have this logic already in `validateEthereumAddress` - can we refactor both these and those functions to avoid code duplication?

packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
```diff
@@ -1,12 +1,19 @@
...
-import { IPersistedAsset } from '../interfaces'
+import { IErc20Metadata, IIrc30Metadata, IPersistedAsset } from '../interfaces'
 import { AssetVerification } from '../types'
-import { getIrc30MetadataFromFoundryOutput } from '../utils/getIrc30MetadataFromFoundryOutput'
+import { getIrc30MetadataFromFoundryOutput } from '../utils'
 
-export async function requestPersistedAsset(tokenId: string): Promise<IPersistedAsset> {
-    const tokenMetadata = await getIrc30MetadataFromFoundryOutput(tokenId)
+export async function requestPersistedAsset(tokenId: string, chainId?: number): Promise<IPersistedAsset | undefined> {
+    let tokenMetadata: IIrc30Metadata | IErc20Metadata | undefined
```
maxwellmattryan, line 11: We should use the `TokenMetadata` type here rather than the union

maxwellmattryan (changes requested):
Have a few requested changes :pray:

packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
begonaalvarezd: an erc-20 token has as ID the smart contract address eg `0x639A647fbe20b6c8ac19E48E2de44ea792c62c5C`, whereas an irc20 token looks like `0x08d5e57a8b2adf771d475eb7c220efbb5ac4e70c40d33c691812c0e63feabe81a80100000000`, so the irc30 token id will fail the ethereum address check on both length and checksum

packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
begonaalvarezd: pushed

begonaalvarezd: I was thinking, do we want to include the base token metadata here then?

packages/shared/lib/core/wallet/actions/getAccountAssetsForSelectedAccount.ts
```diff
@@ -103,28 +101,7 @@ function getAccountAssetForChain(chainId: number): IAccountAssetsPerNetwork | un
                     ...persistedAsset,
                     balance: _balance,
                     chainId,
-                    standard: 'Layer 2 Native Token',
```
begonaalvarezd, line 106: We should also update the check above here 
```
if (persistedAsset && persistedAsset?.metadata && isValidIrc30(persistedAsset.metadata))
```
What makes an erc20 token valid? The ethereum documentation says

> Keep in mind that there are no floats in Solidity. Therefore most tokens adopt 18 decimals and will return the total supply and other results as followed 1000000000000000000 for 1 token

but the decimals could theoretically be up to 255 since its a `uint8`  https://eips.ethereum.org/EIPS/eip-20

Tuditi: Althought theoretically it could be more than 18, I think it's sensible to cap it to 18 here. There is no reason to allow a lot more, since most tokens just copy the default value (18). Additionally, on the Shimmer EVM, only the last 6 decimals are used.

See `Danger` part: https://wiki.iota.org/shimmer/smart-contracts/guide/evm/compatibility/

begonaalvarezd: gotcha, thanks!

packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
maxwellmattryan: Yes, we should

packages/shared/lib/core/token/utils/isValidErc20.ts
```diff
@@ -0,0 +1,6 @@
+import type { TokenMetadata } from '@core/wallet'
+import { MAX_SUPPORTED_ERC20_DECIMALS } from '@core/wallet/constants'
+
+export function isValidErc20(tokenMetadata: TokenMetadata): boolean {
```
maxwellmattryan, line 4: Feels strange to not have token appended to the end of this function name, i.e. `isValidErc20Token`

packages/shared/lib/core/wallet/helpers/buildPersistedAssetFromMetadata.ts
```diff
@@ -0,0 +1,27 @@
+import { NotVerifiedStatus, TokenStandard } from '../enums'
+import { IErc20Metadata, IIrc30Metadata, IPersistedAsset } from '../interfaces'
+import { AssetVerification } from '../types'
+
+export function buildPersistedAssetFromMetadata(
+    tokenId: string,
+    metadata: IIrc30Metadata | IErc20Metadata,
```
maxwellmattryan, line 7: We should use the `TokenMetadata` type here as well

maxwellmattryan (changes requested):
Have a few minor requested changes :pray:

maxwellmattryan (changes requested):
Have a few minor requested changes :pray:

packages/shared/lib/core/wallet/actions/requestPersistedAsset.ts
MarkNerdi: but why though? i think we never need to fetch the metadata for the base token

begonaalvarezd: In this specific case the variable `tokenMetadata` will never be of type `TokenMetadata` as its set by the functions `getErc20TokenMetadata` & `getIrc30MetadataFromFoundryOutput`. Maybe later on we can make something more generic, but for now I would stick to the strict types that we use

packages/shared/lib/core/wallet/helpers/buildPersistedAssetFromMetadata.ts
begonaalvarezd: Same comment as above 🙏🏼

Tuditi (approved):
LGTM

Issue #6690, opened 2023-04-29
[Task]: Create function to transfer ISCP tokens (on L2) for ledger


maxwellmattryan:
### Task description

This involves sending wrapped L1 native tokens and SMR on L2 using the ledger.

### Requirements

_TBD_

### Creation checklist

- [x] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #6934, opened 2023-06-01
feat: allow sending with ledger

Tuditi:
## Summary

closes: #6690 

...

## Changelog

```
- Add functionality to send base token and irc-30 token with ledger.
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

Add the following code to a button e.g. `openCollectibles()` in `Sidebar.svelte`

for base token (1 SMR):
```
const asset = $selectedAccountAssets[1071]?.baseCoin
if (asset) {
    void signIscpTransferTransactionData('0x2ff33407c26e36c32ca50a4e63ce661b2eeed3dd', asset, '1000000')
}
```

for native token (1 unit):

```
const asset = $selectedAccountAssets[1071]?.nativeTokens[0]
if (asset) {
    void signIscpTransferTransactionData('0x2ff33407c26e36c32ca50a4e63ce661b2eeed3dd', asset, '1')
}
```
## Checklist


packages/desktop/components/drawers/network-config/drawers/ConfirmLedgerEvmAddressDrawer.svelte
```diff
@@ -10,7 +10,7 @@
 
     export let drawerRouter: Router<DrawerRoute>
 
-    $: address = $selectedAccount?.evmAddress
+    $: address = $selectedAccount?.evmAddresses[60]
```
MarkNerdi, line 13: can you do the following to remove the magic number?
```suggestion
    $: address = $selectedAccount?.evmAddresses[$selectedChain.cointype]
```

packages/desktop/webpack.config.ts
```diff
@@ -39,6 +39,9 @@ const fallback: { [index: string]: string | false | string[] } = {
     path: false,
     fs: false,
     crypto: false,
+    // These are required for the Ethereum libraries
```
maxwellmattryan, line 42: Would be good to explicitly mention which libraries

packages/shared/lib/core/layer-2/interfaces/layer2-asset-allowance.interface.ts
```diff
@@ -1,7 +1,7 @@
 import { NativeTokenAmount } from '@core/token'
 
 export interface ILayer2AssetAllowance {
-    baseTokenAmount: string
+    baseTokens: string
```
maxwellmattryan, line 4: Seems this name should be singular rather than plural?

packages/shared/lib/core/ledger/interfaces/ledger.interface.ts
```diff
@@ -1,6 +1,6 @@
-import { IEvmTransactionData } from '@core/layer-2/interfaces'
+import type { TxData } from '@ethereumjs/tx'
 
 export interface ILedger {
     generateEvmAddress(bip32Path: string, verify?: boolean): void
-    signEvmTransaction(data: IEvmTransactionData, bip32Path: string): void
+    signEvmTransaction(data: TxData, bip32Path: string): void
```
maxwellmattryan, line 5: Would be good to create a wrapper type around this so we can help slightly with coupling to a particular library - if need to swap the underlying type out it will be easier with the wrapper

packages/shared/lib/core/network/utils/getTxOptions.ts
```diff
@@ -0,0 +1,11 @@
+import { ChainId } from '@core/network'
+import { Common } from '@ethereumjs/common'
+import { TxOptions } from '@ethereumjs/tx'
+
+export function getTxOptions(chainId: ChainId): TxOptions {
```
maxwellmattryan, line 5: - We should not use shortened names / words per the coding conventions, i.e. this should be `getTransactionOptions`
- Would be nice creating a wrapper type around `TxOptions` so that the previous rule isn't broken but mainly it will be easier to change later should we need / want to change the underlying data object we're using.

packages/shared/lib/core/layer-2/utils/getCommonEvmTransactionData.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ISC_MAGIC_CONTRACT_ADDRESS } from '@core/layer-2/constants'
+
+export async function getCommonTransactionData(provider: Web3, originAddress: string, data: string): Promise<TxData> {
+    const nonce = provider.utils.toHex(await provider.eth.getTransactionCount(originAddress))
+
+    const _gasPrice = await provider.eth.getGasPrice()
+    const gasPrice = '0x' + _gasPrice
+
+    const estimatedGas = await provider.eth.estimateGas({ from: originAddress, to: ISC_MAGIC_CONTRACT_ADDRESS, data })
+    const gasLimit = provider.utils.toHex(2 * estimatedGas) // Double to ensure we have enough gas
```
maxwellmattryan, line 12: We have a `GAS_MULTIPLIER` constant that we should use here, avoiding a random magic constant. Perhaps `GAS_LIMIT_MULTIPLIER` is slightly more explicit too?

maxwellmattryan (changes requested):
Requested some changes :pray:

maxwellmattryan (changes requested):
Requested some changes :pray:

packages/shared/lib/core/layer-2/interfaces/layer2-asset-allowance.interface.ts
Tuditi: I agree with you, but this is not the object that the WASP node expects, so we have to request some changes.

Tuditi: It's even uglier for `nativeTokens`, because the ID property expects a string tuple. I'll open an issue in WASP, since they might change this.

packages/shared/lib/core/layer-2/utils/getCommonEvmTransactionData.ts
Tuditi: I will add `GAS_MULTIPLIER`, but we will need to make this whole gas setting dynamic in the future.

packages/shared/lib/core/layer-2/utils/signIscpTransferTransactionData.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function signIscpTransferTransactionData(
+    recipientAddress: string,
+    asset: IAsset,
+    amount: string
+): Promise<void> {
+    try {
+        const chain = getNetwork()?.getChain(asset.chainId)
+        const provider = chain?.getProvider()
+        const evmAddress = getSelectedAccount()?.evmAddresses?.[60]
```
MarkNerdi, line 24: do we have a constant or enum for `60`?

packages/shared/lib/core/network/tests/network.test.ts
```diff
@@ -52,7 +52,7 @@ describe('File: network.ts', () => {
     describe('Function: getOfficialNodes', () => {
         it('should return the correct official nodes given a valid network type', () => {
             Object.values(NetworkId).forEach((networkId) => {
-                if (networkId !== NetworkId.Custom) {
+                if ([NetworkId.Iota, NetworkId.Shimmer].includes(networkId)) {
```
MarkNerdi, line 55: do we also need "Testnet" here?

Tuditi: Yup, I added it :relaxed:

PR #6943, opened 2023-06-02
fix: Ledger device and profile mismatch

maxwellmattryan:
## Summary

In case the user is using the wrong Ledger device or a different seed than the one they made the profile with, we need to protect them against trying to do things like sending, minting, etc.

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

Test with a different device and also using the 25-word phrase.

## Checklist


packages/shared/lib/core/ledger/utils/isMatchingActiveProfile.ts
```diff
@@ -0,0 +1,28 @@
...
+import { LedgerAppName } from '../enums'
+import { isLedgerAppOpen } from './isLedgerAppOpen'
+
+export async function isMatchingActiveProfile(): Promise<boolean | undefined> {
+    if (isLedgerAppOpen(LedgerAppName.Shimmer)) {
+        try {
+            const account = get(selectedAccount)
+            const cachedAddress = account?.depositAddress
+
+            const accountIndex = account?.index ?? get(selectedAccountIndex)
```
MarkNerdi, line 15: either
```suggestion
            const account = get(selectedAccount)
            if (!account)  {
                return
            }
            const cachedAddress = account.depositAddress

            const accountIndex = account.index
```

or 
```suggestion
            const account = get(selectedAccount)
            const cachedAddress = account?.depositAddress

            const accountIndex = get(selectedAccountIndex)
```

MarkNerdi (changes requested):
looks nice! 1 small comment

I think we should add this check on the EVM address generation as well, such that we dont even cache a wrong address

maxwellmattryan:
> I think we should add this check on the EVM address generation as well, such that we dont even cache a wrong address

It is slightly complicated b/c we're using the Shimmer Ledger app to generate an address to check against the cached one we generate when the user creates a profile / account. In the current flow this would mean that the user has to connect Ledger and open the Ethereum app to proceed, then have the Shimmer app opened to generate the address to check that it's matching, then go back to the Ethereum app to actually generate the address. We can't do this with the EVM address because we don't automatically generate it when we create a profile / account. 

I think the better solution would be to change the designs slightly on the previous page such that we include a third Ledger status requirement that the mnemonic's are matching, but even then it requires them have the Shimmer app open first to verify that the device has the right mnemonic, then open the Ethereum app to proceed to the next step where we actually generate the EVM address. I definitely agree that we should prevent the users from caching the wrong address, but perhaps will require a bit more thought on how to do this smoothly in terms of UX.

packages/shared/lib/core/ledger/utils/isLedgerDeviceMatchingActiveProfile.ts
```diff
@@ -0,0 +1,30 @@
...
+import { selectedAccount } from '@core/account/stores'
+import { profileManager } from '@core/profile-manager/stores'
+
+import { LedgerAppName } from '../enums'
+import { isLedgerAppOpen } from './isLedgerAppOpen'
+
+export async function isLedgerDeviceMatchingActiveProfile(): Promise<boolean | undefined> {
+    if (isLedgerAppOpen(LedgerAppName.Shimmer)) {
+        try {
+            const account = get(selectedAccount)
```
Tuditi, line 12: Use `getSelectedAccount()` instead of `get(selectedAccount)` to decouple from svelte stores. Ideally, this should be done for the other functions as well.

Tuditi (changes requested):
Requested some minor changes, maybe we should add this convention to the coding conventions?

Tuditi (approved):
LGTM! :fire:

PR #6990, opened 2023-06-15
Chore/rewrite events to promises

MarkNerdi996:
## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

...

## Checklist


packages/shared/lib/core/layer-2/utils/signIscpTransferTransactionData.ts
```diff
@@ -11,44 +10,44 @@ import {
...
-            Ledger.signEvmTransaction(transaction, bip32)
-        } else {
-            throw new Error('Unable to find web3 provider.')
-        }
-    } catch (err) {
-        console.error(err)
+): Promise<string | undefined> {
+    const chain = getNetwork()?.getChain(asset.chainId)
+    if (!chain) {
+        return
```
Tuditi, line 23: Throw an error here as well?

packages/desktop/electron/lib/ledger.js
```diff
@@ -80,20 +80,23 @@ async function signTransactionData(data, bip32Path) {
...
+            TX_OPTIONS
+        )
+
+        const serializedSignedTransaction = Buffer.from(RLP.encode(bufArrToArr(signedTransactionObject.raw())))
+        const serializedSignedTransactionString = '0x' + serializedSignedTransaction.toString('hex')
+
+        return { signedTransaction: serializedSignedTransactionString }
+    } catch (error) {
+        return { signedTransaction: undefined }
+    }
```
Tuditi, line 101: Try/catching here defeats the purpose of the try/catch statement on l.28 above.

packages/shared/lib/core/layer-2/utils/signTransactionWithLedger.ts
```diff
@@ -0,0 +1,36 @@
...
+        if (!isSigning) {
+            if (signedTransaction) {
+                return Promise.resolve(signedTransaction)
+            } else {
+                return Promise.reject('Rejected')
+            }
+        }
+        await sleep(pollingInterval)
+    }
+    return Promise.reject('Timeout')
```
Tuditi, line 35: This works very nicely. I think we should replace the error messages with locale's. The error on l.30 can say something like: `You must open the Ethereum app on your Ledger device`?

The error on l.35 could use the `error.ledger.temout` locale.

Tuditi (changes requested):

packages/desktop/electron/lib/ledger.js
MarkNerdi996: the use for try catch is that if you `Reject` on the ledger device, we want to return undefined to the renderer process, such that we can handle rejecting there

packages/shared/lib/core/layer-2/utils/signTransactionWithLedger.ts
MarkNerdi996: the error on line 30 will be thrown if the user `Rejects` the signing. So it should rather be `Signing was rejected by the Ledger device`.

Tuditi: It's also thrown when the wrong app is open, but I'll add fixing this as a polish task. Ideally we should use Matt's function to check which app is open and show a pop-up if it's the wrong one.

packages/desktop/electron/lib/ledger.js
Tuditi: Aaah, okay, I see now, where you use this. Maybe we can improve this by sending back the error, so we can make a distinction between the user rejecting or the wrong app being opened on the device. In any case, it's outside the scope of this PR.

Tuditi (approved):
LGTM! :fire:

Issue #7373, opened 2023-08-29
[Task]: Fix `insufficientFunds` error


begonaalvarezd:
### Task description

In the send confirmation popup seems like we have a race condition where `prepareOutput` is being called multiple times due to the component re-mounting and sending right away after coming from the stronghold backup popup

Please note that the issue is present also in `release/desktop-2.1.6`, its not related to "The PR"

![image](https://github.com/iotaledger/firefly/assets/3624944/a97c25e7-dcd5-4327-8b9d-0bd4c803e701)

### Requirements

Fix `insufficientFunds` error

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #7378, opened 2023-08-30
fix: misleading insufficient funds on send

msarcev:
## Summary

> Adds a check so that the `preparedOutput` won't be recomputed after "Stronghold unlocking" step and `sendOutputAndClosePopup` callback execution.

Previously we had the case where, after ` checkActiveProfileAuth` (stronghold unlock step), the `SendConfirmationPopup` component would be re-mounted, also recomputing `preparedOutput` before executing the callback, causing the wallet to throw an error on this subsequent `prepareOutput()` because the output was already "locked" in the previous execution.

Closes https://github.com/iotaledger/firefly/issues/7373

## Changelog

```
- Fixed a misleading insufficient funds error (visible only in dev mode) cause by trying to lock the same output(s) twice 
```

### Platforms

-   **Desktop**
    -   [x] MacOS
    -   [ ] Linux
    -   [ ] Windows
-   **Mobile**
    -   [ ] iOS
    -   [ ] Android

### Instructions

1. Log into a stronghold profile
2. While stronghold is "locked"
3. Send a simple transaction to another wallet (even for the same profile)
4. Unlock stronghold and finish sending
5. You should see an error in console and screen overlay with "insufficient funds" while also having the transaction go through

...

## Checklist

### DIFF
--- a/packages/desktop/components/popups/send/SendConfirmationPopup.svelte
+++ b/packages/desktop/components/popups/send/SendConfirmationPopup.svelte
@@ -25,7 +25,7 @@
     import { DEFAULT_TRANSACTION_OPTIONS } from '@core/wallet/constants'
     import { getOutputParameters, validateSendConfirmation, getAddressFromSubject } from '@core/wallet/utils'
     import { Activity, NewTokenTransactionDetails, Output } from '@core/wallet/types'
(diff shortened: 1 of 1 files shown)


brancoder (approved):
LGTM

PR #7398, opened 2023-09-04
fix: avoid crash on empty NFT details

### LINKED ISSUE #7397: [Task]: Fix empty NFT freezing the app
### Task description

[Example NFT in the explorer](https://explorer.shimmer.network/testnet/addr/rms1zzk7v0tfswwsn385nfcnznd0xgwd0ks48ktv9e3w5z6j5jxpvyzk58sczmj)

image_720

### Requirements

Looking at the details should not freeze/crash the app.

### Creation checklist

VmMad:
## Summary

Fixes #7397 
> Bring back missing optional chaining to avoid crashes.

...

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows

### Instructions

> Mint empty nft with the CLI. Click on the NFT details. It should not crash.

...

## Checklist


packages/shared/lib/core/wallet/utils/getBech32AddressFromAddressTypes.ts
```diff
@@ -2,9 +2,9 @@ import { api } from '@core/profile-manager'
 import { Address, AddressType, AliasAddress, Ed25519Address, NftAddress } from '@iota/sdk/out/types'
 import { getNetworkHrp } from '@core/profile'
 
-export function getBech32AddressFromAddressTypes(address: Address): string | undefined {
+export function getBech32AddressFromAddressTypes(address: Address | undefined): string | undefined {
```
cpl121, line 5: Is it necessary to set this `undefined` here?
I think that `?` is enough

marc2332: The fact that he is using `?` means that the address might be undefined at some point, so it's better that the argument type reflects this case

cpl121 (approved):

packages/shared/lib/core/wallet/utils/getBech32AddressFromAddressTypes.ts
begonaalvarezd, line 5: I think the bugfix for this issue should not come from making this argument optional, as it must be compulsory that you pass an address in order to be able to retrieve its bech32, what needs fixing is all the type errors we have in the code where we are actually calling this function with something that might be undefined, eg
image
We need to make sure in the code that we are not calling this with something that is undefined, and only call it when the types are right

begonaalvarezd (changes requested):

begonaalvarezd (changes requested):
The crash is still here 🙏🏼 
image

begonaalvarezd (approved):
🚀 now it works

PR #7423, opened 2023-09-13
Fix: Add check for Pending Transactions to SendFormPopup

### LINKED ISSUE #7380: [Task]: Add Balance check to improve sending flow logic
### Task description

When attempting to reuse an output in a second transaction while the first transaction is still pending, users may encounter a "Wallet Insufficient Funds" error.
ex.

1. User has consolidated funds to one output and does two transactions in a row
2. Reusing output that holds native tokens in two subsequent transactions.

We should check balance for `baseCoin.available`  before calling `prepareOutput`

Steps to repro:
1. Consolidate funds from multiple outputs into a single output.
2. Create first transaction (base or native tokens)
3.  While first transaction is pending create second transaction

### Requirements

Send two transactions in a row

### Creation checklist

brancoder:
## Summary

Add check for Pending Transactions to `SendFormPopup` to disable sending of multiple transactions in fast manner.
resolves #7380 

### Platforms

-   **Desktop**
    -   [x] MacOS
    -   [x] Linux
    -   [ ] Windows

## Checklist


packages/shared/locales/en.json
```diff
@@ -1751,7 +1751,7 @@
             }
         },
         "syncing": "Please wait until synchronization is finished to change wallets.",
-        "transferring": "Please wait until all transactions are completed to change wallets.",
+        "transferring": "Please wait until all transactions are completed.",
```
msarcev, line 1754: @begonaalvarezd Can we just change this string like this ?
The removed part makes it sound like it is used in some other place for a different purpose. Wdyt ?

brancoder: I've removed the last part because that string is not used anywhere, looks like a leftover so i reused it

begonaalvarezd: i would say its ok, now its more generic

PR #7453, opened 2023-09-25
fix: avoid crash when changing tab in send asset popup

### LINKED ISSUE #7340: [Bug report]: Firefly no response from btn after wrong string entry in nft tab
### Platform

Desktop

### Version

2.1.5

### Description

Wallet crashes when inserting wrong string in nft send tab

### Expected behaviour

no crash

### Actual behaviour

crash

### Can the issue reliably be reproduced?

Yes

Log in the account
click send asset
switch to nft sending
insert receiver address in the nft tag field
go back to send token (because I thought the nft field would be blank again - faster than deleting the string)
go back to nft tab
FF "crashes" can not click on any btn

### Operating System

Windows

### Diagnostics

_No response_

### Developer Profile

Yes

### Profile Type

Software (Stronghold)

### Staking

None

### Crash Status

Non-responsive

### Error Messages

_No response_

### Error Logs

_No response_

### Duplicate declaration

- [X] I have searched the issues tracker this issue and there is none

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

VmMad:
## Summary

Fixes #7340 
> Fix bug in send asset popup to avoid crashes in releases.
...

## Changelog

```
- Fix types in `getNftByIdFromAllAccountNfts` and `NftInput` component
- Only set "selected nft" if the nft exists
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows

### Instructions

- Click send asset
- Switch to "Send nft"
- In NFT Id, insert a random string (or your own address for example) 
- Switch to "Send token" tab
- Switch to "Send nft" tab again. No crash should happen now.

...

## Checklist


marc2332 (approved):
Works :zap:

begonaalvarezd (changes requested):
It does not work properly, if you switch tabs and there was a selected NFT, it should select it again when you get back to the tab, and now it doesnt.
There is a reactivity issue in this implementation and its because before `selected` was initialized directly calling `getNftByIdFromAllAccountNfts` which always retuned an NFT if the id was right, but now selected depends on `selectedNftById` which gets filled reactively, so very likely if `nftId` had a value when the component mounts, `selected` will not see it

begonaalvarezd (approved):

PR #7520, opened 2023-10-02
feat: Vite

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7517

Replace Webpack with Vite

I thought I would be able to first replace the frontend, and then in another PR, do the electron code, but that's not really possible as webpack is hooked in both places, so I just went all in and replaced it in both places

Maybe I'll also replace jest with vitest


marc2332:
Not worth migrating as of now. Rollup is very slow, slower than webpack. Perhaps, when Vite's `rolldown` comes out, we can try again.

PR #7555, opened 2023-10-06
hotfix: Migrate remaining chrysalis profiles

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7556


packages/desktop/views/update-stronghold/views/UpdateStrongholdView.svelte
```diff
@@ -33,8 +35,17 @@
                 await migrateStrongholdFromOnboardingProfile(password)
                 emitStrongholdMigrationEvent({ success: true, onboardingType })
             } else {
-                await migrateStrongholdFromActiveProfile(password)
-                emitStrongholdMigrationEvent({ success: true })
+                try {
+                    await migrateStrongholdFromActiveProfile(password)
+                    emitStrongholdMigrationEvent({ success: true })
+                } catch (err) {
```
begonaalvarezd, line 41: This will eat the error, which we still need for failed migrations that indeed failed for another reason, we need to throw an excepcion for non v3 profiles that still failed 🙏🏼

packages/desktop/views/update-stronghold/views/UpdateStrongholdView.svelte
```diff
@@ -33,8 +35,17 @@
...
+                    const message = err?.message ?? ''
+                    if (message.includes('input snapshot')) {
```
begonaalvarezd, line 43: these 2 strings look a bit generic, is this the best match we can make?

packages/shared/lib/core/profile/actions/profiles/checkAndMigrateChrysalisProfiles.ts
```diff
@@ -152,6 +152,9 @@ function isChrysalisProfile(profile: IPersistedProfile | IChrysalisPersistedProf
         if (chrysalisProfileNetworkId) {
             return chrysalisNetworkIdsArray.includes(chrysalisProfileNetworkId)
         }
+        const hasChrysalisNode =
+            chrysalisProfile.settings?.networkConfig?.nodes?.some((value) => isOfficalChrysalisNode(value.url)) || false
```
begonaalvarezd, line 156: it could be a custom node only, can we maybe asume that if it has any node in this key is actually chrysalis? i dont see this model in stardust so  maybe as soon as an array with an element exists perhaps we can asume is chrysalis

packages/shared/lib/core/profile/constants/default-persisted-profile-object.constant.ts
```diff
@@ -21,6 +22,6 @@ export const DEFAULT_PERSISTED_PROFILE_OBJECT: IPersistedProfile = {
     accountPersistedData: {},
     isDeveloperProfile: false,
     forceAssetRefresh: false,
-    strongholdVersion: undefined,
+    strongholdVersion: StrongholdVersion.V2,
```
begonaalvarezd, line 25: We should not set our default to v2 in my opinion, shoul dbe v3 because all new profiles will be v3

begonaalvarezd (changes requested):

packages/desktop/views/update-stronghold/views/UpdateStrongholdView.svelte
marc2332: No, we will improve this

packages/shared/lib/core/profile/constants/default-persisted-profile-object.constant.ts
marc2332: We should assume the user comes with an V2 as with don't know the exact version

- User profile comes with V2 -> We try to migrate, no error is thrown, profile is migrated and we set the profile to v3
- User profile comes with V3 -> We try to migrate, and error is thrown internally but it's skipped, we simply set the profile to v3

begonaalvarezd: `DEFAULT_PERSISTED_PROFILE_OBJECT` is used during onboarding too, it doesnt make sense doing it this way, we can simply set the version to v3 in the chrysalis to stardust migration instead of affecting the entire app

marc2332: Gotcha, just updated this in https://github.com/iotaledger/firefly/pull/7555/commits/3c3bf612047c1ff524036697813416e005604217, looks good?

begonaalvarezd (approved):

PR #7573, opened 2023-10-10
fix: limit Collectibles name length in Send asset popup

### LINKED ISSUE #7515: [Task]: Limit Collectibles name length in Send asset popup
We should limit the length of the Collectibles names, so they don't overflow like this:
image

evavirseda:
## Summary

> Closes #7515 

...

## Changelog

```
- Truncate key
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows

### Instructions

...

## Checklist

### DIFF
--- a/packages/shared/components/SelectorInput.svelte
+++ b/packages/shared/components/SelectorInput.svelte
@@ -103,7 +103,7 @@
                                 color="gray-800"
                                 classes="truncate"
                             >
-                                {truncateString(option.key, 20, 0) ?? option.value}
+                                {truncateString(option.key, 10, 0) ?? option.value}
                             </Text>
                         </div>
                         {#if option.value && option.key}
--- a/packages/shared/components/SelectorInput.svelte
+++ b/packages/shared/components/SelectorInput.svelte
@@ -103,7 +103,7 @@
                                 color="gray-800"
                                 classes="truncate"
                             >
-                                {option.key ?? option.value}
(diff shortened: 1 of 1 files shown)


begonaalvarezd (changes requested):
I dont think we should add this fixed truncate to the whole selector input which is used in multiple places, the selector is meant to adapt to the with of the container, in this case for example much more than 10 characters fit, and this change is limiting it to 10
image
I would try to approach this with the ellipsis truncate using CSS which is much more flexibke

packages/shared/components/SelectorInput.svelte
```diff
@@ -92,7 +92,12 @@
                         on:click={() => onClick(option)}
                         class="w-full flex flex-row flex-1 justify-between items-center px-2 py-3 rounded-lg hover:bg-blue-50 dark:hover:bg-gray-800 dark:hover:bg-opacity-20"
                     >
-                        <div class="flex flex-row gap-3 justify-start items-center flex-1">
+                        <div
+                            class="flex flex-row gap-3 justify-start items-center flex-1 overflow-hidden {option.value &&
+                            option.key
+                                ? 'max-w-[50%]'
+                                : ''}"
```
begonaalvarezd, line 99: it works for me too without the need of `max-w-[50%]`, do we need it? 
image

begonaalvarezd (changes requested):

PR #7623, opened 2023-10-18
fix: request unlock during login when it has 0 accounts, so we can create a new account

### LINKED ISSUE #7536: [Task]: Allow entering a ledger even though the ledger is not connected
### Task description

At the moment, ledger users cant enter their profiles if the ledger isnot connected, we should allow them

### Requirements

Allow entering a ledger even though the ledger is not connected

### Creation checklist

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7620, Closes #7536

## Reproduce error
1.  Remove `testnet-2` in `getNetworkIdFromNetworkName.ts`
2. Create a profile. It will throw an error at the end, something about a nestwork mismatch, that's fine
3. Enable back `testnet-2` 
4. Try to log into the profile you created in .2
5. It will ask you for the password/ledger, if you click on cancel it will bring you back to the Pin view, if you log in successfully it will proceed to create an account for your profile, and finally log into the profile
6. :v:


cpl121 (approved):
Nice job!!
Works for me 🚀

Issue #7173, opened 2023-07-12
[Task]: Fix L2 transaction metadata decoding for received transactions


brancoder:
### Task description

Parsing (decoding) of metadata for L2 to L1 transactions (transactions recieved from ShimmerEVM) is currently broken.

![image](https://github.com/iotaledger/firefly/assets/52735957/87b00a5c-1154-406e-99e8-2da169de8e18)

### Requirements

Parse metadata so its displayed in the same fashion as on sent transactions.
Requires this PR to be merged: https://github.com/iotaledger/firefly/pull/7124

![image](https://github.com/iotaledger/firefly/assets/52735957/76af9d0a-1dbd-42f9-9d13-dda4985f1d36)

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #7604, opened 2023-10-17
feat: Improved l2 estimated gas

marc2332:
Depends on https://github.com/iotaledger/firefly/pull/7601
Closes https://github.com/iotaledger/firefly/issues/7464

PR #7631, opened 2023-10-19
fix: Update ShimmerEVM txs metadata decoding

marc2332:
Depends on https://github.com/iotaledger/firefly/pull/7604
Closes https://github.com/iotaledger/firefly/issues/7173


cpl121 (approved):
works 🚀

msarcev (approved):

packages/shared/lib/core/layer-2/utils/parseLayer2MetadataForTransferV2.ts
```diff
@@ -4,18 +4,15 @@ import { CONTRACT_FUNCTIONS, TARGET_CONTRACTS } from '../constants'
 import { Allowance } from '../enums'
 import { ReadSpecialStream } from '../classes'
 
-// Function to parse data from the current metadata, using the new encoding where the shimmer chainId is 1072
+// Function to parse data from the current metadata, using the new encoding where the shimmer chainId is 1073
 export function parseLayer2MetadataForTransferV2(metadata: Uint8Array): ILayer2TransferAllowanceMetadata {
```
begonaalvarezd, line 8: This encoding is not v2 anymore but rather v3, but in order to avoid maintaining legacy code forever, i would remove this and just have support with the latest version

begonaalvarezd (changes requested):

begonaalvarezd (approved):

Issue #6280, opened 2023-03-27
[Task]: Update Sentry SDK to latest version


maxwellmattryan:
### Task description

While we are currently using the Sentry SDK, the version is the same as what we initially used in V1, so we should update to the latest version. It's likely that there will be some breaking changes, mainly with where the distribution files are at within the NPM package and also using a specific distribution file for the preload script.

:no_entry_sign: Blocked until #6177 is completed.

### Requirements

- [ ] Update `@sentry/electron` to latest version*
- [ ] Update `@sentry/webpack-plugin` to latest version*
- [ ] Fix existing code and configuration as needed (will likely need to modify `sentry.js` and `preload.js`)

\* It's most likely best to do these separately to see if only one of them causes an issue.

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #7647, opened 2023-10-23
chore: Update Sentry

marc2332:
Closes https://github.com/iotaledger/firefly/issues/6280


packages/desktop/webpack.config.ts
```diff
@@ -189,16 +189,21 @@ const preloadPlugins = [
 ]
 
 const sentryPlugins = [
-    new SentryWebpackPlugin({
-        authToken: process.env.SENTRY_AUTH_TOKEN,
-        include: '.',
-        release: `${getAppName(prod)}@${NETWORK}-${version}`,
-        ignoreFile: '.sentrycliignore',
```
begonaalvarezd, line 196: do we want to stop using this file then? if so, lets remove it, or use it again instead of writing the ignored files directly in the file

begonaalvarezd (changes requested):

packages/desktop/webpack.config.ts
marc2332: Looks like we can only use an array of files now, https://www.npmjs.com/package/@sentry/webpack-plugin#sourcemapsignore

begonaalvarezd (approved):

Issue #7655, opened 2023-10-24
[Task]: Update electron to max possible


begonaalvarezd:
### Task description

Update electron to max possible

### Requirements

Update electron to max possible

### Creation checklist

- [X] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #7675, opened 2023-10-30
chore: Update electron and related dependencies

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7655


patches/electron-updater+5.3.0.patch
```diff
@@ -1,25 +0,0 @@
-diff --git a/node_modules/electron-updater/out/windowsExecutableCodeSignatureVerifier.js b/node_modules/electron-updater/out/windowsExecutableCodeSignatureVerifier.js
```
rajivshah3, line 1: It looks like we'll still need to patch this issue because the latest version of `electron-updater` still has this problem https://github.com/electron-userland/electron-builder/blob/1af7447edf47303de03ca2924727c78118161c60/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L42-L46

packages/desktop/package.json
```diff
@@ -48,9 +48,9 @@
         "copy-webpack-plugin": "^11.0.0",
         "cross-env": "^7.0.2",
         "css-loader": "^6.7.3",
-        "electron": "22.3.27",
-        "electron-builder": "^24.0.0",
-        "electron-notarize": "^1.0.0",
+        "electron": "27.0.2",
+        "electron-builder": "^24.6.4",
```
rajivshah3, line 52: I'm not sure if this was in previous versions of `electron-builder`, but we may need to adjust our `electron-builder-config.js` so that we don't notarize the app twice (it's unclear what the default behavior of the [`notarize` option](https://www.electron.build/configuration/mac) is — it seems to wrap around `@electron/notarize`)

patches/electron-updater+5.3.0.patch
marc2332: > It looks like we'll still need to patch this issue because the latest version of `electron-updater` [still has this problem](https://github.com/electron-userland/electron-builder/blob/1af7447edf47303de03ca2924727c78118161c60/packages/electron-updater/src/windowsExecutableCodeSignatureVerifier.ts#L42-L46)

Do you mind lending me a hand on this one? I am not very familiar with git patches

marc2332: Just added it back

packages/desktop/package.json
marc2332: We weren't signing twice, but now have moved the logic so electron-builder will sign for us

marc2332: @begonaalvarezd fyi

begonaalvarezd (approved):

Issue #7693, opened 2023-11-07
[Task]: Fix `allowance = output amount` l1-l2 tx leads to funds stuck in layer 1.5


begonaalvarezd:
### Task description

We have a user with funds stuck in layer 1.5 because it seems like we either got a 0 gasFee or something, so `allowance = output amount` --> instant fail because there are no base tokens left to pay for the gas fee. We should never allow this case. Note that in that tx `"gasBudget": 9007199254740990` provides a hint that probably the request to wasp to estimate the gas failed

PR #7696, opened 2023-11-07
fix: do not allow sending if failed to request `estimategas-onledger`

begonaalvarezd:
Close https://github.com/iotaledger/firefly/issues/7693

### DIFF
--- a/packages/shared/lib/core/layer-2/utils/getEstimatedGasForTransferFromTransactionDetails.ts
+++ b/packages/shared/lib/core/layer-2/utils/getEstimatedGasForTransferFromTransactionDetails.ts
@@ -1,14 +1,11 @@
-import BigInteger from 'big-integer'
+import { localize } from '@core/i18n'
 import { getActiveProfile } from '@core/profile'
-
-interface GasEstimatePayload {
-    gasBurned?: number
-    gasFeeCharged?: number
-}
+import BigInteger from 'big-integer'
+import { ILayer2GasEstimatePayload } from '../interfaces'
 
 export async function getEstimatedGasForTransferFromTransactionDetails(
     serializedOutputHex: string
-): Promise<GasEstimatePayload> {
+): Promise<ILayer2GasEstimatePayload> {
     const profile = getActiveProfile()
     const chainMetadata = profile.network?.chains?.[0] ?? null
 
@@ -32,8 +29,12 @@ export async function getEstimatedGasForTransferFromTransactionDetails(
             const gasBurned = BigInteger(data.gasBurned as string).toJSNumber()
             const gasFeeCharged = BigInteger(data.gasFeeCharged as string).toJSNumber()
 
-            return { gasBurned, gasFeeCharged }
+            if (gasBurned && gasFeeCharged) {
+                return { gasBurned, gasFeeCharged }
+            }
         }
+
+        throw new Error(localize('error.layer2.estimatedGas'))
     }
 
     return {}
--- a/packages/shared/lib/core/wallet/utils/getOutputParameters.ts
+++ b/packages/shared/lib/core/wallet/utils/getOutputParameters.ts
@@ -1,5 +1,5 @@
 import { getSelectedAccount, prepareOutput } from '@core/account'
-import { localize } from '@core/i18n'
+import { ILayer2GasEstimatePayload } from '@core/layer-2/interfaces'
 import {
     getEstimatedGasForTransferFromTransactionDetails,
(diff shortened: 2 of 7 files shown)


packages/shared/lib/core/wallet/utils/getOutputParameters.ts
```diff
@@ -101,35 +108,48 @@ async function buildOutputParametersForLayer2(transactionDetails: NewTransaction
...
+    } catch (error) {
+        if (error?.message) {
+            showAppNotification({
+                type: 'error',
+                message: error?.message,
+            })
+        }
     }
-
-    return outputParams
```
marc2332, line 132: Did you really want to remove this line?  It will now only return the outputParams if it reaches the L142

packages/shared/lib/core/wallet/utils/getOutputParameters.ts
```diff
@@ -101,35 +108,48 @@ async function buildOutputParametersForLayer2(transactionDetails: NewTransaction
...
+                return outputParams
+            }
+        }
+    } catch (error) {
+        if (error?.message) {
+            showAppNotification({
+                type: 'error',
+                message: error?.message,
+            })
+        }
```
marc2332, line 151: Is it really necessary to catch the error here? I think you can just let it bubble up and it will be catched in `prepareTransactionOutput()`

marc2332 (changes requested):

packages/shared/lib/core/wallet/utils/getOutputParameters.ts
begonaalvarezd: it was intentional yes, we should error or not return anything if there was any issue trying to get the gas estimation

packages/shared/lib/core/layer-2/utils/getEstimatedGasForTransferFromTransactionDetails.ts
```diff
@@ -32,8 +29,12 @@ export async function getEstimatedGasForTransferFromTransactionDetails(
             const gasBurned = BigInteger(data.gasBurned as string).toJSNumber()
             const gasFeeCharged = BigInteger(data.gasFeeCharged as string).toJSNumber()
 
-            return { gasBurned, gasFeeCharged }
+            if (gasBurned && gasFeeCharged) {
+                return { gasBurned, gasFeeCharged }
+            }
         }
+
+        throw new Error(localize('error.layer2.estimatedGas'))
```
cpl121, line 37: If we are going to throw an error, I think it would be a good idea to handle it showing a notification
If not, the application crashes
Captura de pantalla de 2023-11-16 15-13-58

packages/shared/lib/core/wallet/utils/getOutputParameters.ts
```diff
@@ -103,7 +109,7 @@ async function buildOutputParametersForLayer2(transactionDetails: NewTransaction
 
     let estimatedData = await getEstimateData()
 
-    if (estimatedData.gasEstimatePayload.gasBurned) {
+    if (estimatedData?.gasEstimatePayload?.gasBurned) {
```
cpl121, line 112: If there is no `gasBurned`, the `outputParams` are not returned in the `buildOutputParametersForLayer2` function.
As in the previous comment, I think we should handle it and show an error notification, to avoid a crash in the future. In this case, we then try to get the amount of the `outputParams`, but since it is `undefined`, it crashes :trollface: 
Captura de pantalla de 2023-11-16 15-20-09

cpl121 (changes requested):

packages/shared/lib/core/layer-2/utils/getEstimatedGasForTransferFromTransactionDetails.ts
begonaalvarezd: It doesnt crash so dramatically for me 🤔 The error bubbles and its handled later on

begonaalvarezd: ok found it, it was a promise reject that was crashing everything, should be fixed now

packages/shared/lib/core/wallet/utils/getOutputParameters.ts
begonaalvarezd: ok found it, it was a promise reject that was crashing everything, should be fixed now

cpl121 (approved):
LGTM

PR #7702, opened 2023-11-09
fix: l2 metadata decoding

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7694

PR #7703, opened 2023-11-10
fix: shift with BigInt instead of JS number

Tuditi:
## Summary

Bloom team encountered a bug with certain numbers that are decoded as 32-bit numbers instead of 64 numbers due to how JS number bitwise operations work. The solution was to convert all the values to `BigInt()` before executing bitwise operations.

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift#description

...

## Changelog

```
- Bitwise operations on BigInt instead of numbers.
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows

### Instructions

Run tests and try to send 10000999999 glow from L1 to L2.

...

## Checklist


packages/shared/lib/core/layer-2/tests/parseLayer2MetadataForTransfer.test.ts
```diff
@@ -21,6 +21,25 @@ describe('Function: parseLayer2MetadataForTransfer.ts', () => {
         expect(parsedMetadata).toEqual(expected)
     })
 
+    it('should correctly parse metadata with long base token', () => {
```
begonaalvarezd, line 24: In this PR https://github.com/iotaledger/firefly/pull/7702 we already had the bigint fix so this test was added too with some improvements like the ethereum address shouldnt be as large, I think this test can be removed as the one in `develop` is correct already 🙏🏼

begonaalvarezd (changes requested):
Hi @Tuditi thank you so much for submitting this PR ❤️ 
In this PR https://github.com/iotaledger/firefly/pull/7702 we already had the bigint fix but we dont have the code beautified as you do in `packages/shared/lib/core/layer-2/classes/special-stream.class.ts`, if you fix the conflicts we can merge your special stream changes as I see it passes the tests correctly 🌷 
Thanks!

begonaalvarezd (approved):
works 🚀

Issue #7695, opened 2023-11-07
[Task]: Add feature to withdraw base tokens stuck funds in layer 1.5


begonaalvarezd:
### Task description

eg https://archive.evm.shimmer.shimmer.network/v1/chains/smr1prxvwqvwf7nru5q5xvh5thwg54zsm2y4wfnk6yk56hj3exxkg92mx20wl3s/core/accounts/account/smr1qzd26lvs8avvgghrvdk68e3jfd9sjy0emguysz9p8sedda0cv6lhsdjsswu/balance

- Fetch and display layer 1.5 base token balance when the base token is selected 
![image](https://github.com/iotaledger/firefly/assets/3624944/7759a648-5c94-4465-8163-b815d422ad20)
- Add a withdrawal option if there are funds
- Withdrawal popup: display the balance & withdrawal logic (wasp request)

PR #7734, opened 2023-11-20
feat: allow l2 withdrawal (funds stuck in l2 that didnt reach evm)

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7695


cpl121 (changes requested):
I'm trying the PR, but I think it's not working correctly. I create a stuck tx, and then i try to withdraw L2 funds, but I have a `Not enough balance.` error and the `Withdraw from L2` popup stays in an infinite loop

brancoder:
> I'm trying the PR, but I think it's not working correctly. I create a stuck tx, and then i try to withdraw L2 funds, but I have a `Not enough balance.` error and the `Withdraw from L2` popup stays in an infinite loop

What is the amount that you tried to withdraw?

packages/desktop/components/popups/WithdrawFromL2Popup.svelte
```diff
@@ -0,0 +1,170 @@
...
+    const { isStrongholdLocked } = $activeProfile
+
+    $: password && address && !$isStrongholdLocked && withdrawFromL2()
+
+    function onCancelClick(): void {
+        closePopup()
+    }
+
+    function onWithdrawFromL2Click(): void {
+        openUnlockStrongholdPopup()
```
cpl121, line 40:
```suggestion
        if ($isSoftwareProfile && $isStrongholdLocked) {
        openUnlockStrongholdPopup()
        } 
```
We would only have to unlock the stronghold, if it is previously locked

cpl121 (changes requested):

packages/desktop/components/popups/WithdrawFromL2Popup.svelte
```diff
@@ -0,0 +1,197 @@
...
+                bip44Chain,
+                gasEstimate.toString()
+            )
+        } else {
+            isWithdrawing = false
+            showErrorNotification(localize('error.send.notEnoughBalance'))
+            return
+        }
+
+        try {
```
cpl121, line 106: In function `getLayer2WithdrawRequest`, we need to accept de signature with the ledger device but the verification popup does not open

1. This `try catch` should be placed just above the first call to the `getLayer2WithdrawRequest` function so that if the tx is canceled the `withdraw` button does not remain in an infinite loop
2. I don't know if we can open the verification popup directly, but if not we could put a notification like this at least 
`if ($isActiveLedgerProfile && $ledgerNanoStatus.connected) {
                    displayNotificationForLedgerProfile('info', null, null, 'Look your ledger device to sign the transaction')
                }`
wdyt?

cpl121 (changes requested):
Nice job!
It works correctly with the software profiles, but I have found a problem with ledger

cpl121 (approved):
works perfectly for both profiles (ledger and software) 🚀 
Amazing job @brancoder 👏🏼

begonaalvarezd (approved):
🔥🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

Issue #7801, opened 2023-12-21
[Task]: Fix hardcoded password during onboarding


marc2332:
- Requires to make a PR to iota-sdk so `set_stronghold_password` is available for SecretManager in nodejs
- Remove harcoded passwords
- Clean up code

PR #7804, opened 2023-12-21
refactor: Dynamic password

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7801
~Requires https://github.com/iotaledger/iota-sdk/pull/1807~


marc2332:
@begonaalvarezd we can now merge this

cpl121 (approved):
Works!

packages/shared/lib/contexts/onboarding/stores/onboarding-secret-manager.store.ts
```diff
@@ -1,25 +1,41 @@
...
+
+        // 4. Store Mnemonic
+        await secretManager.storeMnemonic(mnemonicStringified)
+
+        // 5. Generate address
+        const address = (await secretManager.generateEd25519Addresses({}))[0]
+
+        updateOnboardingProfile({
+            address,
+        })
```
begonaalvarezd, line 38: the address does not seem to be necessary, the wallet can be initialized with the secret manager and the correct bip path, and then the SDK should be able to generate the address by itself

marc2332: Just removed the address creation and instead I will now pass the password to the wallet options the first time the wallet is created

begonaalvarezd (approved):

PR #7923, opened 2024-01-23
feat: set & persist explicit account created from the implicit flow as the selected one

### LINKED ISSUE #7845: [Task]: Set & persist explicit account created from the implicit flow as the selected one
### Task description

The created account should the one be used as the selected account to be used for everything account related, like issue the blocks, use as remainder address, use as deposit address, accumulate potential mana converted to stored mana, ... so we need to persisit the reference to this account so we can use it later on, the SDK will not store which one to use from the list

evavirseda:
## Summary

> Closes #7845 

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [x] Linux
    -   [ ] Windows

### Instructions

...

## Checklist


packages/desktop/views/dashboard/wallet/views/WalletView.svelte
```diff
@@ -1,8 +1,16 @@
 <script lang="ts">
     import { AssetList, Overflow, Pane, ReceiveAddressButton } from '@ui'
     import { AccountSummary, AccountActivity, SendButton } from '@components'
-    import { selectedWalletAssets } from '@core/wallet'
+    import { mainAccountId, selectedWalletId, selectedWalletAssets } from '@core/wallet'
     import features from '@features/features'
+    import { onMount } from 'svelte'
+    import { updateActiveWalletPersistedData } from '@core/profile'
+
+    onMount(() => {
```
cpl121, line 9: I think this is not correct. I would have to update it 3 times
- When the account is created for the first time (handleEventNewOutput)
- When the accounts exist and there is no value (buildWalletState)
- When it can be changed in account management (account management)

packages/shared/lib/core/wallet/stores/main-account-id.store.ts
```diff
@@ -0,0 +1,7 @@
+import { writable } from 'svelte/store'
+
+export const mainAccountId = writable<string>(null)
```
cpl121, line 3:
```suggestion
export const mainAccountId = writable<string | null>(null)
```

packages/shared/lib/core/wallet/utils/getDepositAddress.ts
```diff
@@ -1,12 +1,14 @@
...
 import { IWallet } from '../../profile/interfaces'
 import { getBech32AddressFromAddressTypes } from './getBech32AddressFromAddressTypes'
+import { updateMainAccountId } from '../stores'
 
 // TODO(2.0): Update the implementation to handle multiple accounts and use the selected account's address.
 export async function getDepositAddress(wallet: IWallet): Promise<string> {
     const accounts = await wallet?.accounts()
     if (accounts?.length) {
         const accountId = (accounts[0]?.output as AccountOutput).accountId
+        updateMainAccountId(accountId)
```
cpl121, line 11: the same as the first comment, I think it is not the place to update de main account id

packages/shared/lib/core/wallet/stores/main-account-id.store.ts
```diff
@@ -0,0 +1,7 @@
...
+
+export function updateMainAccountId(payload: string): void {
+    mainAccountId.update(() => payload)
```
cpl121, line 6:
```suggestion
    mainAccountId.set(payload)
```

cpl121 (changes requested):

packages/shared/lib/core/wallet/actions/buildWalletState.ts
```diff
@@ -42,6 +43,10 @@ export async function buildWalletState(
         accountOutputs = await wallet.accounts()
         implicitAccountOutputs = await wallet.implicitAccounts()
         walletOutputs = await wallet.outputs()
+        const accountId = (accountOutputs[0]?.output as AccountOutput).accountId
+        if (accountOutputs.length) {
+            updateMainAccountId(accountId)
+        }
```
cpl121, line 49:
```suggestion
        if (accountOutputs.length && !mainAccountId) {
            const accountId = (accountOutputs[0]?.output as AccountOutput).accountId
            updateMainAccountId(accountId)
        }
```

packages/shared/lib/core/wallet/actions/events-handlers/handleNewOutputEvent.ts
```diff
@@ -68,6 +70,10 @@ export async function handleNewOutputEventInternal(walletId: string, payload: Ne
     if (isAccountOutput) {
         const accounts = await wallet.accounts()
         const depositAddress = await getDepositAddress(wallet)
+        const accountId = (accounts[0]?.output as AccountOutput).accountId
+        if (accounts) {
+            updateMainAccountId(accountId)
+        }
         updateSelectedWallet({
             accountOutputs: accounts,
             depositAddress,
```
cpl121, line 79:
```suggestion
        if (accounts.length === 1) {
            const accountId = (accounts[0]?.output as AccountOutput).accountId
            updateMainAccountId(accountId)
            const depositAddress = await getDepositAddress(wallet)
             updateSelectedWallet({
                    accountOutputs: accounts,
                    depositAddress,
             })
        } else {
              updateSelectedWallet({
                    accountOutputs: accounts,
              })
        }
```
```

**Review by cpl121 — CHANGES_REQUESTED** [review]
On the other hand, the implementation of `getDepositAddress` would also have to be updated

File: packages/shared/lib/core/wallet/stores/main-account-id.store.ts [hunk 7]
```diff
@@ -0,0 +1,13 @@
+import { get, writable } from 'svelte/store'
+import { updateActiveWalletPersistedData } from '../../profile'
+import { selectedWalletId } from './selected-wallet-id.store'
+
+export const mainAccountId = writable<string | null>(null)
+
+export function updateMainAccountId(payload: string): void {
+    mainAccountId.set(payload)
+    // Update persisted data
+    updateActiveWalletPersistedData(get(selectedWalletId), {
```
cpl121, line 10: something weird is happening here I think
Captura de pantalla de 2024-01-24 12-19-27
This creates a wallet with null id

cpl121 (changes requested):

cpl121 (approved):
Perfectt! NJ 🚀

packages/shared/lib/core/wallet/actions/buildWalletState.ts
```diff
@@ -42,6 +44,10 @@ export async function buildWalletState(
         accountOutputs = await wallet.accounts()
         implicitAccountOutputs = await wallet.implicitAccounts()
         walletOutputs = await wallet.outputs()
+        if (accountOutputs.length && !get(mainAccountId)) {
```
begonaalvarezd, line 47: why do we set the main account when we build the wallet?

packages/shared/lib/core/wallet/actions/events-handlers/handleNewOutputEvent.ts
```diff
@@ -66,11 +68,19 @@ export async function handleNewOutputEventInternal(walletId: string, payload: Ne
 
     if (isAccountOutput) {
         const accounts = await wallet.accounts()
-        const depositAddress = await getDepositAddress(wallet)
-        updateSelectedWallet({
-            accountOutputs: accounts,
-            depositAddress,
-        })
+        if (accounts.length === 1) {
+            const accountId = (accounts[0]?.output as AccountOutput).accountId
```
begonaalvarezd, line 72: what does the SDK return when we call `wallet.accounts()`, is it all account outputs, or only block issuer accounts? 
I ask because if they return all, in order to be 100% accurate, we should check that this output has the block issuer feature enabled (for which we can make a util function to check if an account is a block issuer) otherwise we could potentially be setting as main account an account that cant issue blocks

packages/shared/lib/core/wallet/stores/main-account-id.store.ts
```diff
@@ -0,0 +1,13 @@
+import { get, writable } from 'svelte/store'
+import { updateActiveWalletPersistedData } from '../../profile'
+import { selectedWalletId } from './selected-wallet-id.store'
+
+export const mainAccountId = writable<string | null>(null)
```
begonaalvarezd, line 5: will this store load its initial value when the wallet loads, taking the value that is stored in the local storage? For me this value should be a derived store that is derived from the selected wallet

packages/shared/lib/core/wallet/stores/main-account-id.store.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function updateMainAccountId(payload: string, walletId = get(selectedWalletId)): void {
```
begonaalvarezd, line 7: we shouldnt be setting this store explicitly, but rather just updating the persisted data and that propagates to our store

packages/shared/lib/core/wallet/utils/getDepositAddress.ts
```diff
@@ -1,15 +1,16 @@
...
 import { IWallet } from '../../profile/interfaces'
 import { getBech32AddressFromAddressTypes } from './getBech32AddressFromAddressTypes'
+import { get } from 'svelte/store'
+import { mainAccountId } from '../stores'
 
-// TODO(2.0): Update the implementation to handle multiple accounts and use the selected account's address.
 export async function getDepositAddress(wallet: IWallet): Promise<string> {
     const accounts = await wallet?.accounts()
-    if (accounts?.length) {
+    if (accounts?.length && !get(mainAccountId)) {
```
begonaalvarezd, line 9: there should not be deposit address if there is no `mainAccount`

begonaalvarezd (changes requested):

packages/shared/lib/core/wallet/actions/buildWalletState.ts
cpl121: Only set if there are accounts and no previous main account. When we do login, this is where the wallet is built, so we think it's the right place to set a default main account
Maybe we could also check that the account has a feature for block issuer and if not, not set any by default. wdyt?

packages/shared/lib/core/wallet/actions/events-handlers/handleNewOutputEvent.ts
cpl121: Yes, `accounts` returns all outputs, whether or not they have the issuer feature block or not. This is what the sdk function filter does
`Checks whether the output is a(n) AccountOutput`

packages/shared/lib/core/wallet/actions/buildWalletState.ts
begonaalvarezd: alright then, yes we should always check before setting an account that the account indeed has the block issuer feature, so first we have to know if the SDK returns with `accounts` only accounts that can issue blocks or not

cpl121: The sdk returns all the accounts, we will have to do the feature block issuer filter

begonaalvarezd: lets do some utils out of it then, you can try some kind of `getBlockIssuerAccounts(wallet)` using a `hasBlockIssuerFeature(account)` or `isBlockIsserAccount(account)` that you can use here and there

cpl121 (changes requested):
amazing job! 🥇 
2 more comments
1. I think it would be a good idea in `ImplicitAccountCreationView` to check `if ($selectedWallet?.hasImplicitAccountCreationTransactionInProgress && $selectedWallet?.isTransferring)` to redirect to step 3. This way, when we receive more extra funds from the faucet, it is not sent to step 2 again
2. The wallet deposit address does not update properly when the account output is received, and looks empty
Captura de pantalla de 2024-01-30 12-31-42

packages/shared/lib/core/wallet/actions/buildWalletState.ts
```diff
@@ -29,31 +31,44 @@ export async function buildWalletState(
...
         implicitAccountOutputs = await wallet.implicitAccounts()
         walletOutputs = await wallet.outputs()
+
+        // initialize selectedWalletMainAccountId if there is none set so the wallet can be used
+        // TODO: check that selectedWalletMainAccountId is still an owned account
+        const _selectedWalletMainAccountId = get(selectedWalletMainAccountId)
+        const blockIssuerAccounts = await getBlockIssuerAccounts(wallet)
+
+        // check if the current selectedWalletMainAccountId is still owned by the wallet
+        if (
```
cpl121, line 52: I don't know if this is the place to handle this. I think this should be updated when an account output is spent, not when the wallet is being built

packages/shared/lib/core/profile/actions/active-profile/login.ts
```diff
@@ -56,6 +56,11 @@ export async function login(loginOptions?: ILoginOptions): Promise<void> {
                 initialSelectedWalletId = lastUsedWalletId
             }
             setSelectedWallet(initialSelectedWalletId)
+            const _selectedWallet = get(selectedWallet)
```
cpl121, line 59: add comment here about why we do this here instead of in `buildWalletState` 😄

packages/shared/lib/core/wallet/utils/getDepositAddress.ts
```diff
@@ -1,13 +1,20 @@
...
-// TODO(2.0): Update the implementation to handle multiple accounts and use the selected account's address.
 export async function getDepositAddress(wallet: IWallet): Promise<string> {
-    const accounts = await wallet?.accounts()
-    if (accounts?.length) {
-        const accountId = (accounts[0]?.output as AccountOutput).accountId
+    const accountOutputsWithBlockIssuerFeature = await getBlockIssuerAccounts(wallet)
+    const _selectedWalletMainAccountId = get(selectedWalletMainAccountId)
+    if (_selectedWalletMainAccountId) {
+        const accountAddress = getBech32AddressFromAddressTypes(new AccountAddress(_selectedWalletMainAccountId))
+        return accountAddress ?? ''
```
cpl121, line 13:
```suggestion
        return accountAddress
```

cpl121 (changes requested):
so close!! 🤏🏼

cpl121 (approved):
Cool!

packages/shared/lib/core/wallet/utils/getDepositAddress.ts
```diff
@@ -1,14 +1,21 @@
...
+import { selectedWalletMainAccountId, updateSelectedWalletMainAccountId } from '../stores'
+import { getBlockIssuerAccounts } from './outputs'
 
-// TODO(2.0): Update the implementation to handle multiple accounts and use the selected account's address.
 export async function getDepositAddress(wallet: IWallet): Promise<string> {
-    const accounts = await wallet?.accounts()
-    if (accounts?.length) {
-        const accountId = (accounts[0]?.output as AccountOutput).accountId
+    const accountOutputsWithBlockIssuerFeature = await getBlockIssuerAccounts(wallet)
+    const _selectedWalletMainAccountId = get(selectedWalletMainAccountId)
```
begonaalvarezd, line 10: the deposit address should not dependend on the selected wallet, should be agnostic, as it was before

packages/shared/lib/core/profile/actions/active-profile/login.ts
```diff
@@ -57,6 +65,28 @@ export async function login(loginOptions?: ILoginOptions): Promise<void> {
             }
             setSelectedWallet(initialSelectedWalletId)
 
+            // we need to initialize selectedWallet first and then get the deposit address, otherwise the deposit address will be empty
+            const _selectedWallet = get(selectedWallet)
```
begonaalvarezd, line 69: the logic of the deposit address should not be in login, we should try to keep it where it was before, in the build wallet logic

packages/shared/lib/core/profile/actions/active-profile/login.ts
```diff
@@ -57,6 +65,28 @@ export async function login(loginOptions?: ILoginOptions): Promise<void> {
...
+            const blockIssuerAccounts = await getBlockIssuerAccounts(_selectedWallet)
+
+            // check if the current selectedWalletMainAccountId is still owned by the wallet
+            if (
+                _selectedWalletMainAccountId &&
+                !blockIssuerAccounts.find(
+                    (account) => (account?.output as AccountOutput)?.accountId === _selectedWalletMainAccountId
+                )
+            ) {
+                updateSelectedWalletMainAccountId(undefined)
```
begonaalvarezd, line 87: this logic should likely be moved out of here too

begonaalvarezd (changes requested):

packages/shared/lib/core/wallet/actions/events-handlers/handleNewOutputEvent.ts
```diff
@@ -65,12 +69,28 @@ export async function handleNewOutputEventInternal(walletId: string, payload: Ne
...
+            !get(selectedWalletMainAccountId) &&
+            wallet?.hasImplicitAccountCreationTransactionInProgress &&
+            hasBlockIssuerFeature(accountOutput)
+        ) {
+            updateSelectedWalletMainAccountId(accountOutput.accountId)
+            updateActiveWallet(walletId, {
+                hasImplicitAccountCreationTransactionInProgress: false,
+                isTransferring: false,
+            })
+        }
```
cpl121, line 93: All this logic also needs to be updated 🙏🏼

cpl121 (changes requested):

cpl121 (approved):
Perfect!

begonaalvarezd (approved):

Issue #7830, opened 2023-12-23
[Task]: Cleanup imports


begonaalvarezd:
### Task description

There are ugly imports after the refactor that point to eg shared/lib/core/... instead of use the alias or relative paths for modules that should import within it using relative paths

PR #8008, opened 2024-02-07
chore: Clean up imports

marc2332:
Closes https://github.com/iotaledger/firefly/issues/7830

- Convert imports pointing to `shared` to the corresponding `@whatever` path
- Moved subrouters from `packages/desktop` to `packages/shared` so we don't need to use absolute paths when importing routers from `packages/desktop in `packages/shared`
- Converted paths with `../../` pointing to a core module to `@core/whatever`
- Moved some components from `packages/desktop` to `packages/shared` so they can be imported from `packages/shared`


evavirseda (commented):
could also update the `AssetList.svelte`??  :pray: 
 
`import { Filter } from '@components'`

marc2332:
> could also update the `AssetList.svelte`?? 🙏
> 
> `import { Filter } from '@components'`

The problem here is that Filter is a component from `packages/desktop` and AssetList is from `packages/shared`. Currently, there is no way to import in a fancy way things from desktop into shared. Let me see what I can do

evavirseda (approved):
nice job!!

begonaalvarezd (approved):
beautiful

Issue #8013, opened 2024-02-08
[Task]: Profiles wallets created in the same session are duplicated


evavirseda:
when importing a profile from a mnemonic, when attempting to import another or create a new one, the previously created profile reappears, causing all profiles to be identical

![Image](https://github.com/iotaledger/firefly/assets/72078237/b89b2c63-f6a5-4834-a601-26cfedf98ef3)

![Image](https://github.com/iotaledger/firefly/assets/72078237/14e7ee7b-8910-4e08-ac4c-f09263f61e8c)

PR #8014, opened 2024-02-08
fix: Prevent duplication of wallets across profiles created in same session

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8013

### DIFF
--- a/packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
+++ b/packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
@@ -7,9 +7,7 @@ export function convertOnboardingProfileToPersistedProfile(
     onboardingProfile: Partial<IOnboardingProfile>
 ): IPersistedProfile {
     return {
-        ...DEFAULT_PERSISTED_PROFILE_OBJECT,
-        walletPersistedData: {},
-        clientOptions: {},
+        ...structuredClone(DEFAULT_PERSISTED_PROFILE_OBJECT),
         ...(onboardingProfile?.id && { id: onboardingProfile.id }),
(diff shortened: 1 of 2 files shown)


packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
```diff
@@ -8,6 +8,8 @@ export function convertOnboardingProfileToPersistedProfile(
 ): IPersistedProfile {
     return {
         ...DEFAULT_PERSISTED_PROFILE_OBJECT,
+        walletPersistedData: {},
```
cpl121, line 11: Why clean the information of all the wallets that are already persisted in the profile?

cpl121 (changes requested):
I don't understand why remove too `lastUsedWalletId`

packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
marc2332: What profile? This is code is for when you create an onboardingWallet, there is nothing persisted

marc2332:
> I don't understand why remove too `lastUsedWalletId`

because of the same reason, this is an onboarding wallet, there is no lastUsedWalletId, it's pointless

packages/shared/lib/contexts/onboarding/helpers/convertOnboardingProfileToPersistedProfile.ts
cpl121: makes sense :trollface:

cpl121 (approved):

begonaalvarezd (approved):

Issue #8033, opened 2024-02-13
[Task]: Maintain depositAddress when making a transaction


cpl121:
### Task description

When making a transaction, the output is spent, so the `depositAddress` is removed and is not updated again when the new output arrives with the account

PR #8024, opened 2024-02-09
feat: succesfully issue a block with base token

cpl121:
Closes #7976 

## Summary

> Please summarize your changes, describing **what** they are and **why** they were made.

...

## Changelog

```
- Please write a list of changes
```

### Platforms

> Please select any platforms where your changes have been tested.

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows

### Instructions

> Please describe the specific instructions, configurations, and/or test cases necessary to **test and verify** that your changes work as intended.

...

## Checklist

> Please tick the following boxes that are relevant to your changes.

-   [ ] I have followed the contribution guidelines for this project
-   [ ] I have performed a self-review of my own code
-   [ ] I have commented my code, particularly in hard-to-understand areas
-   [ ] I have added or modified tests that prove my changes work as intended
-   [ ] I have verified that new and existing unit tests pass locally with my changes
-   [ ] I have verified that my latest changes pass CI workflows for testing and linting
-   [ ] I have made corresponding changes to the documentation

PR #8043, opened 2024-02-14
fix: Maintain deposit address when making txs

marc2332:
Depends on https://github.com/iotaledger/firefly/pull/8024
Closes https://github.com/iotaledger/firefly/issues/8033

Only unselect the active account when its output is being transferred.


packages/shared/lib/core/wallet/actions/events-handlers/handleSpentOutputEvent.ts
```diff
@@ -32,14 +32,14 @@ export async function handleSpentOutputEventInternal(walletId: string, payload:
...
         updateActiveWallet(walletId, { walletOutputs, accountOutputs, implicitAccountOutputs })
-        if (
-            wallet.mainAccountId &&
-            !walletOutputs.find(
-                (output) =>
-                    output.output.type === OutputType.Account &&
-                    (output as unknown as AccountOutput).accountId === wallet.mainAccountId
-            )
-        ) {
+        const isSendingTheMainAccount = walletOutputs.find(
```
begonaalvarezd, line 35: this code here just means that you have an account that has the same ID as your main. We have to check where we can know if the account was spent & returned to you or just spent and send somwhere else or burned. We need to stufy if in this event `walletOutputs` or `accountOutputs` already has the new unspent account ot not. If not, we might no be able to handle this logic here, maybe it could be moved `handleTransactionInclusionEvent` though we should study well the events

begonaalvarezd (changes requested):

cpl121 (approved):

Issue #8159, opened 2024-03-07
[Task]: Fix flagged NFTs check


marc2332:
### Task description

https://github.com/iotaledger/firefly/pull/8025/files#diff-157365104b7c831ac44805cc74b89313c18bb33d30d585fc2f3920ccb9d7eff9

PR #8161, opened 2024-03-07
fix: Update regex for flagged NFTs

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8159

This will flag any text that contains two words joined by a `.`, e.g : `you are special, enter my website: surprise.com` or also some of the keywords included, e.g `congrats, you won the lottery`


marc2332:
We need a better algo and update the i18

packages/shared/locales/en.json
```diff
@@ -2126,6 +2126,9 @@
         },
         "participation": {
             "noFunds": "You do not have any IOTA."
+        },
+        "nft": {
+            "flagged": "Be careful when following unknown links. Never share your private keys, nor enter them into any websites or services.    "
```
begonaalvarezd, line 2131:
```suggestion
            "flagged": "Be careful when following unknown links. Never share your private keys, nor enter them into any websites or services."
```

begonaalvarezd (approved):

Issue #8239, opened 2024-03-21
[Task]: Fix tx history: Receive funds to the implicit account address


begonaalvarezd:
### Task description

Fix tx history so we have a working MVP: Receive funds to the implicit account address(es)

PR #8260, opened 2024-03-25
fix: Receive funds to implicit account address

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8239


cpl121 (approved):

evavirseda (commented):
it's not reactive; the transaction isn't being displayed in the activity column until we log out

image
image

begonaalvarezd (approved):

Issue #8309, opened 2024-04-04
[Task]: Fix async claimed error in activities generation


marc2332:
![image](https://github.com/iotaledger/firefly/assets/38158676/fd69bf06-fcbc-4e66-91c8-3417bf755844)

PR #8317, opened 2024-04-04
fix: Properly send metadata feature in Send Flow

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8309


cpl121 (approved):
Works! 🚀 
Captura de pantalla de 2024-04-04 23-33-02

cpl121:
> Do we really want to support multiple metadata keys?
Yes, I think that makes sense. Other platforms can create multiple metadata entries and we should show it
@marc2332

marc2332:
> > Do we really want to support multiple metadata keys?
> 
> Yes, I think that makes sense. Other platforms can create multiple metadata entries and we should show it @marc2332

Alright, I have unlinked the task so it can be addressed in the future.

evavirseda (approved):
nice job :raised_hands:

Issue #8326, opened 2024-04-05
[Task]: Fix `Wallet was destroyed` crash


begonaalvarezd:
### Task description

When I leave firefly open for a while without doing anything, i usually stumble upon this bug. Now it happened to me when I tried to delete a profile
![image](https://github.com/iotaledger/firefly/assets/3624944/a0a2e6f7-61f7-4623-8bb3-0197b5669bf0)

PR #8338, opened 2024-04-09
fix: Prevent wallet destroyed error

marc2332:
~Depends on https://github.com/iotaledger/firefly/pull/8401~
Closes https://github.com/iotaledger/firefly/issues/8326


packages/shared/lib/core/wallet/actions/getClient.ts
```diff
@@ -2,7 +2,11 @@ import { Client } from '@iota/sdk/out/client'
 import { selectedWalletId } from '@core/wallet/stores'
 import { api } from '@core/api'
 import { get } from 'svelte/store'
+import { getWalletById } from '../../profile'
 
 export function getClient(walletId: string = get(selectedWalletId)): Promise<Client> {
-    return api.getClientFromWallet(walletId)
+    const wallet = getWalletById(walletId)
```
evavirseda, line 8: just curious, but i don't understand why we need to check `wallet` :thinking:  :pray:

packages/shared/lib/core/wallet/actions/syncBalance.ts
```diff
@@ -7,15 +7,17 @@ import { OutputData } from '@iota/sdk/out/types'
...
-    if (syncCongestion) {
-        const wallet = getWalletById(walletId)
-        accountOutputs = wallet?.accountOutputs ?? []
-    }
-    const totalBalance: IBalance = await getTotalWalletBalance(walletId, accountOutputs)
-    if (get(selectedWalletId) === walletId) {
-        updateSelectedWallet({ balances: totalBalance })
-    } else {
-        updateActiveWallet(walletId, { balances: totalBalance })
+    const wallet = getWalletById(walletId)
```
evavirseda, line 10: same in here :pray:

packages/shared/lib/core/wallet/actions/getClient.ts
marc2332: Just to prevent any race condition between the destruction of the wallet and any background polling that we have on firefly, it's just in case.

evavirseda (approved):
LGTM!

cpl121 (approved):
I can't reproduce the error in the base branch, but the code seems fine to me to avoid race conditions

PR #8357, opened 2024-04-10
fix: available mana without counting the implicit

### LINKED ISSUE #8352: [Task]: fix available mana without counting the implicit
### Task description

The available mana returned by the SDK counts all outputs (implicit and explicit counts), unlike base tokens which do filter them

cpl121:
Closes #8352 

## Summary

...

## Changelog

```
```

### Platforms

-   **Desktop**
    -   [ ] MacOS
    -   [ ] Linux
    -   [ ] Windows

### Instructions

...

## Checklist


packages/shared/lib/core/wallet/actions/getTotalWalletBalance.ts
```diff
@@ -1,13 +1,23 @@
...
 import { getBalance } from './getBalance'
+import { getManaBalance, getTotalAvailableMana } from '../../network'
+import { getWalletById } from '../../profile'
 
 export async function getTotalWalletBalance(walletId: string, accountOutputs: OutputData[]): Promise<IBalance> {
     const balances = await getBalance(walletId)
     const blockIssuanceCredits = await getBicBalance(walletId, accountOutputs)
     const totalWalletBic = Object.values(blockIssuanceCredits).reduce((acc, bic) => acc + Number(bic), 0)
+
+    const _selectedWallet = getWalletById(walletId)
```
marc2332, line 12: This is going to break onboarding I think, `getTotalWalletBalance` is called in buildWalletState.ts`

marc2332 (changes requested):

packages/shared/lib/core/wallet/actions/getTotalWalletBalance.ts
cpl121: I haven't had any problems

packages/shared/lib/core/wallet/actions/getTotalWalletBalance.ts
```diff
@@ -1,13 +1,23 @@
...
+    let realAvailableMana = getManaBalance(balances?.mana?.available)
+    if (_selectedWallet) {
```
marc2332, line 14: This doesn't make sense, you still want to get the real available mana when the wallet is created in the onboarding. Imo, `getTotalWalletBalance` should be receiving the wallet instance, not the wallet id. Wallet ID were a mistake to be used as function param.

marc2332: There is absolutely no case where if you call getTotalWalletBalance, `_selectedWallet` should not exist, and if it didn't then the problem is elsewhere. Which in this case is that hasn't been loaded when you are in the onboarding, hence why we should be using wallet instances and not wallet ids.

marc2332 (changes requested):

packages/shared/lib/core/wallet/actions/getTotalWalletBalance.ts
marc2332: This will be avoided in https://github.com/iotaledger/firefly/pull/8404 anyway

marc2332 (approved):

PR #8439, opened 2024-04-24
fix: Always remove active profile on logout

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8429


evavirseda (approved):
:rocket:

cpl121 (approved):

Issue #8194, opened 2024-03-12
[Task]: reactivity error when mining an nft


cpl121:
### Task description

Reactivity error when mining an nft. 1 of the minted nfts is not displayed correctly in the collectibles view

PR #8450, opened 2024-04-25
fix: Mint nfts reactivity

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8194

### DIFF
--- a/packages/shared/lib/core/wallet/actions/events-handlers/handleTransactionInclusionEvent.ts
+++ b/packages/shared/lib/core/wallet/actions/events-handlers/handleTransactionInclusionEvent.ts
@@ -8,7 +8,6 @@ import {
 } from '@iota/sdk/out/types'
 import { updateParticipationOverview } from '@contexts/governance/stores'
 import { isWalletVoting } from '@contexts/governance/utils/isWalletVoting'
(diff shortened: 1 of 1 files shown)


evavirseda (commented):
i don't understand the change... i'm not sure about this solution.. :see_no_evil: :pray:

packages/shared/lib/core/wallet/actions/events-handlers/handleTransactionInclusionEvent.ts
```diff
@@ -58,8 +58,7 @@ export function handleTransactionInclusionEventInternal(
 
     if (activity?.type === ActivityType.Nft) {
         const isSpendable =
-            (activity.direction === ActivityDirection.Incoming ||
-                activity.direction === ActivityDirection.SelfTransaction) &&
+            (activity.direction === ActivityDirection.Incoming || activity.direction === ActivityDirection.Outgoing) &&
```
begonaalvarezd, line 61: I dont think this logic is correct, why would an NFT be spendable by you if you send it out?

begonaalvarezd (changes requested):

begonaalvarezd (approved):

PR #8459, opened 2024-04-25
fix: Wrong time comparison

marc2332:
Closes https://github.com/iotaledger/firefly/issues/8458

### DIFF
--- a/packages/shared/lib/core/wallet/utils/transactions/getSenderAddressFromInputs.ts
+++ b/packages/shared/lib/core/wallet/utils/transactions/getSenderAddressFromInputs.ts
@@ -8,28 +8,24 @@ import {
     UnlockConditionType,
     AccountAddress,
 } from '@iota/sdk/out/types'
-import { getUnixTimestampFromNodeInfoAndSlotIndex, nodeInfoProtocolParameters } from '@core/network'
-import { get } from 'svelte/store'
 
 export function getSenderAddressFromInputs(inputs: IWrappedOutput[]): string | undefined {
     for (const input of inputs) {
         const { output, metadata } = input
         const unlockConditions = (output as CommonOutput)?.unlockConditions
 
-        const nodeParamenters = get(nodeInfoProtocolParameters)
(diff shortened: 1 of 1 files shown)


cpl121 (approved):

packages/shared/lib/core/wallet/utils/transactions/getSenderAddressFromInputs.ts
```diff
@@ -8,28 +8,24 @@ import {
...
-                const expirationUnlockCondition = unlockConditions.find(
-                    (unlockCondition) =>
-                        unlockCondition.type === UnlockConditionType.Expiration &&
-                        (unlockCondition as ExpirationUnlockCondition).slot < spentDate
-                ) as ExpirationUnlockCondition
-
-                if (expirationUnlockCondition) {
-                    return AddressConverter.addressToBech32(expirationUnlockCondition.returnAddress)
-                }
+        const spentSlot = metadata?.spent?.slot
```
begonaalvarezd, line 17: idk if we can rely on the spent slot to determine who signed an output at a given tx, if the block failed an output could be tried to be sent in multiple blocks and only succeed in 1 and the `metadata?.spent?.slot` would be the same for all, but they the same input potentially could be signed by different people at different times. In order to be accurate here, you need the slot of the transaction imo, not the slot of then the input was spent

packages/shared/lib/core/wallet/utils/transactions/getSenderAddressFromInputs.ts
```diff
@@ -8,28 +8,24 @@ import {
...
+
+        if (spentSlot) {
+            // A transaction with an expiration unlock condition is included if the transaction expired
+            const expirationUnlockCondition = unlockConditions.find(
```
begonaalvarezd, line 21: I would rename this to `expiredExpirationUnlockCondition` so its easier to understand

begonaalvarezd (changes requested):

begonaalvarezd (approved):

PR #8668, opened 2025-04-25
chore: updated windows code signing

shufps:
Uses external code signing server using FIPS compliant secure hardware


begonaalvarezd (approved):

marc2332 (approved):
lgtm

